PR #7, opened 2021-02-17
Nicer tooltips for time and age plots

### LINKED ISSUE #3: Improve tooltips in the time and age distribution plots
That's how it currently looks like:

image

A suggestion for improvement:

- The date tooltip could print: "Week 1, 2021 (04.01)"
- Instead of "72 | trace 0": "Number of sequences: 72"
- Instead of "11.46497 | trace 1": "Proportion: 11.46%"

This would also make it much clearer what the lines and bars mean.

tehwalris:
This fixes #3. @chaoran-chen I roughly followed your suggestion. Here are the new tooltips (age plot is the same):

2021-02-17-120126_11520x2160_scrot

I also made the X-axis in the time plot explicitly show weeks, so that it matches the tooltip. This calculates ISO 8601 weeks, instead of directly using the week numbers from the response, because that is all that the formatting system supports. Is that the correct week calculation?

### DIFF
--- a/src/widgets/VariantAgeDistributionPlot.tsx
+++ b/src/widgets/VariantAgeDistributionPlot.tsx
@@ -64,6 +64,9 @@ export const VariantAgeDistributionPlot = ({ data }: Props) => {
           ]}
           layout={{
             title: 'Age Distribution',
+            xaxis: {
+              title: 'Age',
+            },
             yaxis: {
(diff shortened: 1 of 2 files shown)


chaoran-chen (approved):
Yes, that matches very well! Thank you!!

PR #9, opened 2021-02-17
Sample details tooltip

### LINKED ISSUE #4: Sample details tooltip
In the sample list view, when the user hovers over the GISAID ID...

image

...more details about the sample should be shown in a tooltip **if the user is logged in**.

Example content for the tooltip:

> #### EPI_ISL_751193
> **Submitting lab:** Department of Biosystems Science and Engineering, ETH Zürich
> **Country:** Switzerland
> **Division:** Basel-Stadt
> **Location:** 4058 Basel
> **Date:** 2020-12-18
> \---------------------------
> **Host:** Human
> **Age:** 30
> **Sex:** Male

Corresponding API: https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md#sample

tehwalris:
This closes #4. The tooltip format is roughly as you described. If the API does not return metadata, then no tooltips appear (that is what happens without a logged in user). 

If you want any changes to how the tooltips look or behave, please tell me. Otherwise just merge directly.

For now tooltips are "good enough", but I think in the future it would be nice to have a table with all this data in hidable, filterable and summarized (eg. "Country: 80% Switzerland") columns.

### DIFF
--- a/src/components/SampleTable.tsx
+++ b/src/components/SampleTable.tsx
@@ -1,21 +1,54 @@
 import React, { useState, useEffect } from 'react';
-import { Col, Container, OverlayTrigger, Row } from 'react-bootstrap';
-import { OverlayChildren } from 'react-bootstrap/esm/Overlay';
-import Table from 'react-bootstrap/Table';
+import { Col, Container, Overlay, Popover, Row, Table } from 'react-bootstrap';
 
 import { Variant, Country } from '../helpers/types';
 
 import { getSamples } from '../services/api';
-
-//Get information about a sample
-
-// export class SampleTable extends React.Component {
 
 interface Sample {
   name: string;
   date: string;
   country: string;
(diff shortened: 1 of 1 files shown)


src/components/SampleTable.tsx
```diff
@@ -1,27 +1,63 @@
...
+  metadata: SampleMetadata | null;
+}
+
+interface SampleMetadata {
+  age: number; // 0 if unknown
+  country: Country | null;
+  division: string | null;
+  host: string | null;
+  location: string | null;
+  sex: string | null; // is "?" sometimes
```
chaoran-chen, line 22: I know.. I have to clean up the data

chaoran-chen (approved):
Looks great, thanks!

And I agree. In the future, a more powerful data table solution might be better.

PR #11, opened 2021-02-17
API response validation and types

tehwalris:
@TKGZ You should be reviewer, but I can't pick you from the list on GitHub for some reason.

I've added validation for API responses, so that we know what we got from the server actually matches our TS types. This might prevent some hard to find bugs in the future. 

I'm sorry that this PR is a bit big. I don't expect you to look through it line by line, but I want to make sure you're on board with the general changes. Here's what they are:

- Added shema objects to perform the actual validation (you can call `.parse(randomData)` on them) in `services/api-types.ts`
- Created named plain-object TypeScript types from the schemas at the bottom of `services/api-types.ts`
- Adjusted imports for types to come from `services/api-types.ts` instead of `helpers/types.d.ts`
  - This does not include `DataDistributionConfiguration` since that's our internal object, not from the API
  - The types for entries in the "distribution" arrays are slightly renamed (eg `VariantAgeDistributionDataPoint` -> `AgeDistributionEntry`) - we could use the old names if you prefer, I don't mind either way
- Updated `api.js` to work with all of this (convert to TS, add type annotations, call `.parse(...)` on schema objects)

The data flowing through the application hasn't changed. The only minor exception to this is `age` and `sex` in `Sample` where some invalid values are removed (see `.transform(v => (v === '?' ? null : v))`).

I've tried out all the functionality in the app and I don't think that I've broken anything. 

What do you think of these general changes?


spacephoton:
@tehwalris seems like everything runs fine! we can merge :)

Issue #12, opened 2021-02-17
Smart ticks for week axis


tehwalris:
Most of our plots have a time axis in weeks. I used `tickvals` so that there is a tick every week, which looks nice on most time scales. Without this Plotly chooses to place ticks very sparsely and at pretty random intervals.

The problem is that a tick for every week is too much if the plot contains lots of weeks (the labels become vertical and can even overlap). In addition every week is labeled with a year which uses tons of space, but the year only interesting for the boundary between two years, as well as maybe first and last data points of the whole plot.

Ideally we would:
- Show week ticks at uniform intervals
- Show one tick every week as long as that fits well
- Remove ticks "smartly" if we don't have space (keep first week, last week, and weeks near year changes)
- Hide the year on most ticks (show on first week, last week, and near year changes)

We might be able to find some good existing code to do this. Maybe there's a function for spacing weekly ticks in D3 or a related library.

chaoran-chen:
Outdated - we don't use Plotly so much.

PR #13, opened 2021-02-17
Nicer tooltips for international comparison

### LINKED ISSUE #8: Improve tooltips in the international comparison plot
And another tooltip issue...

image

- The x-axis has again a weekly scale, i.e., we would like to replace "Dec 14" with "Week XX, 2020 (from 14.12)"
- Could we have only one tooltip per country? The text could be in this case: "0.55% [0.19%, 1.61%] | Switzerland"

tehwalris:
Closes #8 as described there. I also cleaned up the trace names (they used to have `(trace 0)` at the end). The week labels are not ideal - I opened issue #12 for that. 

2021-02-17-203912_11520x2160_scrot

### DIFF
--- a/src/services/api-types.ts
+++ b/src/services/api-types.ts
@@ -105,6 +105,7 @@ export const LoginResponseSchema = z.object({
 
 // TypeScript types from schemas
 
+export type ValueWithCI = z.infer<typeof ValueWithCISchema>;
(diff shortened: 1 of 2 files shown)


chaoran-chen (approved):
Nice! Thank you!

PR #18, opened 2021-02-20
Add Nextclade integration

### LINKED ISSUE #17: Nextclade integration
Next to the "Show samples" buttons, a new button "Show on Nextclade" should be added. On click, a new tab will be opened redirecting to Nextclade.

The following should happen in the app:

1. Create a temporary JWT token (with `/internal/create-temporary-jwt`)
2. The sequences in fasta format can be fetched with `GET /resource/sample-fasta?<params>&jwt=<temporary token>`
3. Open the following link in a new tab `https://clades.nextstrain.org/?input-fasta=<endpoint URL>`. The endpoint URL has to be encoded (-> use encodeURIComponent()).

The button should be only available for logged-in users.

chaoran-chen:
This closes #17.

### DIFF
--- a/src/components/InternationalComparison.tsx
+++ b/src/components/InternationalComparison.tsx
@@ -7,6 +7,8 @@ import { VariantInternationalComparisonPlotWidget } from '../widgets/VariantInte
 import { DistributionType, getVariantDistributionData } from '../services/api';
 import { Country, InternationalTimeDistributionEntry, Variant } from '../services/api-types';
 import { getSamplePageLink } from '../pages/SamplePage';
+import { AccountService } from '../services/AccountService';
+import { NextcladeService } from '../services/NextcladeService';
 
 interface Props {
   country?: Country;
@@ -84,6 +86,16 @@ export const InternationalComparison = ({ country, matchPercentage, variant }: P
       <div style={{ display: 'flex' }}>
         <h3 style={{ flexGrow: 1 }}>{variant.name ?? 'Unnamed Variant'} - International Comparison</h3>
         <div>
+          {AccountService.isLoggedIn() && (
+            <Button
+              onClick={() => NextcladeService.showVariantOnNextclade(variant, matchPercentage, undefined)}
+              variant='outline-dark'
+              size='sm'
+              className='mr-2'
+            >
+              Show on Nextclade
+            </Button>
+          )}
           <Link to={getSamplePageLink({ mutations: variant.mutations, matchPercentage })}>
             <Button variant='outline-dark' size='sm'>
               Show all samples
@@ -125,6 +137,18 @@ export const InternationalComparison = ({ country, matchPercentage, variant }: P
                     <td>{c.first.yearWeek}</td>
                     <td>{c.last.yearWeek}</td>
                     <td>
+                      {AccountService.isLoggedIn() && (
+                        <Button
+                          onClick={() =>
+                            NextcladeService.showVariantOnNextclade(variant, matchPercentage, c.country)
+                          }
+                          variant='outline-dark'
+                          size='sm'
+                          className='mr-2'
+                        >
+                          Show on Nextclade
+                        </Button>
+                      )}
                       <Link
                         to={getSamplePageLink({
                           mutations: variant.mutations,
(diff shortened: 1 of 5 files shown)


tehwalris (changes requested):
I have tested this and it seems to work fine, including loading in Nextclade. I had to disable mixed content blocking (HTTP/HTTPS) in Chrome, but this won't be an issue once we have the public server with HTTPS.

However I looked at the changed code, including the new API endpoints, and noticed that there's a security issue. If you get access to a Nextclade link, you can abuse it to make **any** private requests to our API **forever**. I understand that ability to make any request with this JWT is probably intentional and simplifies the implementation. I think that the "forever" part was not intentional though:

The API changes added the `/internal/create-temporary-jwt` endpoint, which gives you a JWT token with a short expiry. You need to be logged in (have a valid JWT token) to access this endpoint. The problem is that the `/internal/create-temporary-jwt` can be abused to extend temporary tokens forever. You can call it with a **temporary** JWT token to get another temporary JWT token, but now with a later expiry time. If you can get a recent Nextclade link, you can extend the JWT like this, and use it to access private data (including metadata) for as long as you want.

I see a few solutions ranging from most simple (first) to most secure (last):
- _Do nothing_. I think that's not in line with our goals. We are adding logins and things to protect private data, so we should fix issues that we find with them.
- _Add a `temporary: true` field to temporary JWTs and don't allow them to be used for `create-temporary-jwt`._ I can't estimate how hard this will be in Spring, but it's probably reasonable. This will fix the direct problem.
- _Use JWTs to sign a specific request for a FASTA file._ Instead of issuing a JWT using `create-temporary-jwt` make a `/internal/create-signed-fasta-url` which takes a query (country, mutations, etc) and returns a ready URL which will give a FASTA file with no extra credentials. This would replace `getSampleFastaUrl`. However we can make this put a special JWT in the returned URL which is not bound to a user, but instead **contains** the rest of the URL. This JWT will only let you make a request with exactly the original query. This mechanism is very similar to presigned URLs in S3. This restricts the damage of a accidentally shared URL as much as possible.

I hope my security concerns are reasonable, but please tell me if this is the wrong level of caution for this project.

src/components/InternationalComparison.tsx
```diff
@@ -84,6 +86,16 @@ export const InternationalComparison = ({ country, matchPercentage, variant }: P
       <div style={{ display: 'flex' }}>
         <h3 style={{ flexGrow: 1 }}>{variant.name ?? 'Unnamed Variant'} - International Comparison</h3>
         <div>
+          {AccountService.isLoggedIn() && (
+            <Button
+              onClick={() => NextcladeService.showVariantOnNextclade(variant, matchPercentage, undefined)}
+              variant='outline-dark'
+              size='sm'
+              style={{ marginRight: '10px' }}
```
tehwalris: You can also use `className="mr-1"` (or `mr-2`, [see here](https://getbootstrap.com/docs/5.0/utilities/spacing/)) I know it's not much nicer, but if we switch to a Bootstrap theme with different spacing, then the spacing that these utility classes make will adjust automatically.

chaoran-chen:
Hi @tehwalris, thanks for bringing this up! I've now restricted a temporary token to a single endpoint that has to be defined during its creation. See: https://github.com/cevo-public/cov-spectrum-server/commit/5f5039bb4ff7d4fd2da240b34bf90de578fb07eb

(I initially had only some complicated approaches in mind to implement temporary tokens with limited scope but your third suggestion inspired me.)

tehwalris (approved):
@chaoran-chen I looked at the server changes and I think the solution you've chosen solves the problem well. I haven't tested that it prevents calling `create-temporary-jwt` because I think that the server is not up to date yet. I assume you've tested it though, so from my side it's fine to merge.

A suggestion for a minor server change: I think it would be good to make `restrictionEndpoint` required. There is no use case for a temporary token with no restrictions, right? This would prevent us from accidentally creating one in the future.

chaoran-chen:
Oh, that's a documentation mistake. It is already required!

Only a token with `restrictionEndpoint=/internal/create-temporary-jwt` can be used to create new tokens.

(And yes, I need to make a final change until CD really works.)

PR #22, opened 2021-02-21
Feature/postalmap

### LINKED ISSUE #1: Switzerland Postal Code Map
We would like to have a map that shows the geographical distribution of samples of a selected variant. In this initial step, the map shall present the total number of found samples per zip code-area. It should be a heat map, i.e., the more cases an area had, the darker should its color be.

The map should only be shown if (1) the user is logged in and (2) the selected country is Switzerland.

Corresponding API: https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md#variant-time-zip-code-distribution

spacephoton:


spacephoton:
@tehwalris  are you familiar with topo json map data? I’m trying to use react-simple-maps to render the data in my feature/postalmap branch. When I pass in the swiss-maps1.json file, it works fine, but all the other data with the zip codes as id/or plz do not work, and make browser crash. Any idea what it could be? Alternative approaches? Coulnd’t find anything online that had things include so it is taking longer than expected. @chaoran-chen

chaoran-chen:
Please squash the commits at the end as we don't need hundreds of thousands of lines of data in our git history :)

spacephoton:
there is a 'width' which defines the width of the map, what's the simplest way of getting width of the parent component? @tehwalris

tehwalris:
> there is a 'width' which defines the width of the map, what's the simplest way of getting width of the parent component? @tehwalris

We don't have a system in place yet to supply you a width. However the latest widget code will wrap your component in a fixed height div when used in the main UI and a full-frame div when embedded. You can use `viewbox`for your SVG instead of `width` and `height`, and then you can scale it like an image (width 100%, height 100%). You may have to use some other attributes - there's very good [information here](https://css-tricks.com/scale-svg/#the-preserveaspectratio-attribute). Note with this setup your map won't always be fill the whole width, but I think that's a good enough solution for now.

spacephoton:
> > there is a 'width' which defines the width of the map, what's the simplest way of getting width of the parent component? @tehwalris
> 
> We don't have a system in place yet to supply you a width. However the latest widget code will wrap your component in a fixed height div when used in the main UI and a full-frame div when embedded. You can use `viewbox`for your SVG instead of `width` and `height`, and then you can scale it like an image (width 100%, height 100%). You may have to use some other attributes - there's very good [information here](https://css-tricks.com/scale-svg/#the-preserveaspectratio-attribute). Note with this setup your map won't always be fill the whole width, but I think that's a good enough solution for now.

Yeah, the map scales automatically given the width, but it will stretch beyond the parent's wrapper, so I need to pass the width of parent

tehwalris:
> > > there is a 'width' which defines the width of the map, what's the simplest way of getting width of the parent component? @tehwalris
> > 
> > 
> > We don't have a system in place yet to supply you a width. However the latest widget code will wrap your component in a fixed height div when used in the main UI and a full-frame div when embedded. You can use `viewbox`for your SVG instead of `width` and `height`, and then you can scale it like an image (width 100%, height 100%). You may have to use some other attributes - there's very good [information here](https://css-tricks.com/scale-svg/#the-preserveaspectratio-attribute). Note with this setup your map won't always be fill the whole width, but I think that's a good enough solution for now.
> 
> Yeah, the map scales automatically given the width, but it will stretch beyond the parent's wrapper, so I need to pass the width of parent

Yes, I understand that that's what will currently happen. Like I said, we can't pass the width or height of the parent at the moment.

The easiest way to work around this is to scale your map like an image, by using `viewbox` and `height: 100%'. This can be done without knowing the width/height of the parent in JS. That's what the above article describes.

If you don't want to use this "SVG like an image" technique, then the next best thing is using something to measure your component (eg. [this hook](https://yarnpkg.com/package/react-cool-dimensions)).

Long term we should pass sizes from the parent, since it's the most reliable way. We can plan this out once some other aspects of our application are stabilized.

spacephoton:
> If you don't want to use this "SVG like an image" technique, then the next best thing is using something to measure your component (eg. [this hook](https://yarnpkg.com/package/react-cool-dimensions)).

Ah, I see what you mean. Cool, will look into this tomorrow! Thanks for the link! 
:D

spacephoton:
@tehwalris @chaoran-chen Ready for review. Please let me know if it runs too slow and what errors you might see/what should be fixed. If not, we can merge. 

Couldn't make a scalable SVG while keeping the tooltips (great article though), but found another way!

How do I run prettier with the proper config to pass the test cases here?

chaoran-chen:
You can run prettier with `npm run format`.

spacephoton:
> You can run prettier with `npm run format`.

awesome! done

src/App.tsx
```diff
@@ -69,7 +64,7 @@ export const FocusWrapper = styled.div`
 `;
 
 export const App = () => {
-  const [selection, setSelection] = useState<Selection>();
+  const [selection, setSelection] = useState<Selection | undefined>(undefined);
```
tehwalris, line 67: This is identical to the original line. When you use `useState` without passing any arguments, the default value is `undefined`.

src/components/KnownVariantsList.tsx
```diff
@@ -43,18 +44,29 @@ export const KnownVariantsList = ({ country, onVariantSelect }: Props) => {
...
-                onClick={() => {
-                  onVariantSelect(d);
-                }}
-                variant='outline-secondary'
-                size='sm'
-              >
-                Show Details
-              </Button>
-            </td>
+            {selection !== undefined && selection.variant.name === d.name ? (
```
tehwalris, line 47: This is a useful feature. It's a bit random in this PR, but still nice to have. The table sometimes shifts around slightly when the selection changes. I think this is fine for now - the known variants UI will change later anyway.

src/components/Switzerland/Map.tsx
```diff
@@ -0,0 +1,89 @@
...
+const Wrapper = styled.div`
+  padding: ${PADDING_SCALE}rem ${PADDING_SCALE}rem ${PADDING_SCALE}rem 1rem;
+`;
+
+type Props = {
+  width: number;
+  distributionData: TimeZipCodeDistributionEntry[];
+};
+
+const Map = ({ width, distributionData }: Props) => {
```
tehwalris, line 17: Nice choice on where to separate the different components.

src/components/Switzerland/Map.tsx
```diff
@@ -0,0 +1,89 @@
...
+      </div>
+      <ReactTooltip />
+    </Wrapper>
+  );
+};
+
+export const isEqual = (prevProps: Props, nextProps: Props) =>
+  prevProps.width === nextProps.width && prevProps.distributionData === nextProps.distributionData;
+
+export default React.memo(Map, isEqual);
```
tehwalris: Great that you used `memo` here. I think it would do the same thing without `isEqual` though: "By default it will only shallowly compare complex objects in the props object" ([doc](https://reactjs.org/docs/react-api.html#reactmemo)).

src/components/Switzerland/Map.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import bbox from '@turf/bbox';
+import relief from './relief.jpg';
+import geoJson from './PLZ10.json';
+
+const PADDING_SCALE = 1;
+const WIDTH_ADJUST = 16 * 2 * PADDING_SCALE;
+
+const Wrapper = styled.div`
+  padding: ${PADDING_SCALE}rem ${PADDING_SCALE}rem ${PADDING_SCALE}rem 1rem;
```
tehwalris: You are adding padding in `Map`, but doing measurement in `Switzerland`. I think it will simplify the code if you add the padding in the same component where you are measuring.

Specifically, if you add padding in `Switzerland` instead of `Map`, then you can get rid of most of the padding code (`WIDTH_ADJUST`, `position: relative`, etc). The trick is that `ResizeObserver` (which `react-resize-detector` uses) can measure `content-box` (actually this is the default; [doc](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/observe)). So if you wrap your `Map` in a `div` in `Switzerland`, put the padding on that `div`, and measure it, then the measurements will be without padding.

src/components/Switzerland/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { TimeZipCodeDistributionEntry } from '../../services/api-types';
+import { AccountService } from '../../services/AccountService';
+import * as zod from 'zod';
+import { SampleSelectorSchema } from '../../helpers/sample-selector';
+import { useResizeDetector } from 'react-resize-detector';
+
+import Map from './Map';
+
+const PropsSchema = SampleSelectorSchema;
+type Props = zod.infer<typeof PropsSchema> & { width?: number };
```
tehwalris: `& { width?: number };` is not used anymore.

src/components/Switzerland/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const Switzerland = ({ country, mutations, matchPercentage }: Props) => {
+  const [distributionData, setDistributionData] = useState<TimeZipCodeDistributionEntry[]>([]);
+  const loggedIn = AccountService.isLoggedIn();
```
tehwalris, line 23: I think this login check makes more sense in `FocusPage`.

tehwalris: Ah, I just saw that this component also shows "Please log in" - now this makes more sense. However I think it's probably better just to hide the map compeletely (and not show a message) if the user is logged out, since very few users will ever be able to log in. What do you think @chaoran-chen?

src/components/Switzerland/Map.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useEffect, useState, useRef } from 'react';
+import { geoPath, geoTransform } from 'd3-geo';
+import { scaleLinear } from 'd3-scale';
+import ReactTooltip from 'react-tooltip';
+import { TimeZipCodeDistributionEntry } from '../../services/api-types';
+import { scaleQuantile } from 'd3-scale';
+import styled from 'styled-components';
+
+import bbox from '@turf/bbox';
+import relief from './relief.jpg';
```
tehwalris, line 9: When I first saw `relief.jpg` in the code I thought that it's pretty unecessary. I've seen it in the UI now though, and I think it looks really good. It's the only really nice looking thing in our application so far.

src/components/Switzerland/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+      isSubscribed = false;
+      controller.abort();
+    };
+  }, [country, mutations, matchPercentage]);
+
+  return loggedIn && distributionData !== undefined ? (
+    <>
+      <div ref={ref as React.MutableRefObject<HTMLInputElement>}>
+        <p>Number of cases by postal code (PLZ)</p>
+        <Map width={width ?? 1000} distributionData={distributionData} />
```
tehwalris: I think it's better not to render the map before you have a `width` from `useResizeDetector`. The rendering (in JS, React and browser) is pretty expensive, and it will almost always happen twice, since the measured width usually won't be 1000.

src/services/api-types.ts
```diff
@@ -99,13 +99,13 @@ export const InternationalTimeDistributionEntrySchema = zod.object({
...
-  x: zod
-    .object({
-      week: YearWeekWithDaySchema,
-      zip_code: zod.string(),
-    })
-    .transform(v => ({ week: v.week, zipCode: v.zip_code })),
-  y: CountAndProportionWithCISchema,
+  x: zod.object({
+    week: YearWeekWithDaySchema,
+    zipCode: zod.string(),
```
tehwalris, line 104: @chaoran-chen It looks like the documentation is wrong [here](https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md#variant-time-zip-code-distribution).

src/services/api-types.ts
```diff
@@ -121,7 +121,10 @@ export const GrowingVariantSchema = zod.object({
 export const LoginResponseSchema = zod.object({
   token: zod.string(),
 });
-
+export interface Selection {
```
tehwalris, line 124: There is already a type for something similar in `sample-selector.ts`. You can ues `zod.infer<typeof SampleSelectorSchema>` to get it. We could name it (`type SampleSelector = ...`) in that file if you want.

For your specific usage though (showing the selected variant), I think it nicer to just pass `selectedVariantName` as `string | undefined`, since that's all the components is using anyway.

tehwalris (changes requested):
The map works really well! I haven't seen any bugs or performance issues while testing it.

The code is nice. I only left some notes and small suggestions. Fell free to merge directly once you consider those. BTW You'll also have to fix lint warnings before merging, since those are errors in CI.

chaoran-chen:
I didn't look at the code but tested the feature and the map looks great! Just one suggestion: I think that it would be better to have the map above the "International Comparison" section where we have the super long table and when you don't place your mouse correctly, you might have to scroll through the whole table before you can scroll the page to get to the map.

src/components/Switzerland/index.tsx
chaoran-chen: Later we should take it out but right now, I like the message so that other users know that our tool can do even more amazing things.

src/services/api-types.ts
chaoran-chen: Thanks for pointing it out! Is fixed.

spacephoton:
> I didn't look at the code but tested the feature and the map looks great! Just one suggestion: I think that it would be better to have the map above the "International Comparison" section where we have the super long table and when you don't place your mouse correctly, you might have to scroll through the whole table before you can scroll the page to get to the map.

Yes, I also noticed I would scroll in the table. @tehwalris what do you think?

src/components/Switzerland/index.tsx
spacephoton: - [x] Good point, I will fix

src/App.tsx
spacephoton: I think I added it explicitly to remove some TS errors that were showing up

src/components/Switzerland/Map.tsx
spacephoton: Thanks!
- [x] Remove equal function

src/components/Switzerland/Map.tsx
spacephoton: - [x] Move padding to object where measurement is being done

src/components/Switzerland/index.tsx
spacephoton: - [ ] Remove

spacephoton:
Will fix from comments and merge tomorrow then :)

src/services/api-types.ts
spacephoton: Do you mean simply pass the name instead of the 'Selection' object?

PR #36, opened 2021-03-05
Routing system with bookmarkable URLs

### LINKED ISSUE #29: Routing system / bookmarkable URLs for the variant page
Country and variant selection should change the URL, and vice versa, the user should be able to open a certain country and variant with a direct link.

A first proposal:

```
## Basic pages:
/login
/about

## The splitted explore/focus page:
/e/{country}/ -> No variant selected
/e/{country}/variant?name=B.1.1.7&matchPercentage=0.8 -> known variant
/e/{country}/variant?mutations=....&matchPercentage=0.8 -> not a known variant

## Deep focus:
/e/{country}/variant?mutations=....&matchPercentage=0.8/samples -> sample list
/e/{country}/variant?mutations=....&matchPercentage=0.8/demographics
...
```

The "e" in `/e/` may be understood as an abbreviation for "explore."

tehwalris:
This gives our pages proper URLs (closes #29) and makes other related changes.

I went with nearly the scheme that Chaoran suggested. Here are some example URLs:
```
/about - The about page
/explore/:country - The split explore-focus page with no variant selected
/explore/:country/variants/:variantSelector - The split explore-focus page with a variant selected
/explore/:country/variants/:variantSelector/samples - The deep focus page showing detailed sample information
```
The `:variantSelector` is encoded using the same system that the "Show samples" page used (`encodeURIComponent(JSON.stringify(data))`). The encoded selector is not pretty, but the code is reliable and simple.

The only thing that doesn't fit into this URL scheme is the page with the table of samples. The problem is that it is possible to show that page without a country. I decided to split it into 2 pages (`/global-samples/:variantSelector` and `/explore/:country/variants/:variantSelector/samples`). This solves lots of minor problems (eg. you shouln't show the country selection box when showing worldwide samples). There's almost no code duplication since both pages use `SampleTable`. By using `LazySampleButton` you always get links to the correct page (depending on if you pass a country or not). Note that the global page is only linked from "Show worldwide samples" in "International comparison".

Since the samples screen (with country) is under `/explore/:country/variants/:variantSelector/samples`, it is now our first "deep focus" page. I added some very basic styling for that, roughly like in the mockups. I think we should put move the tables from "International comparison" to a deep focus page too - I'll open an issue for that.

@chaoran-chen and @TKGZ Try it out and tell me if anything in the UI is confusing or broken. Feedback about the URL scheme and layout of the new "deep focus" pages is also welcome.

@TKGZ You can choose how carefully you want to review the code. I know it's not very practical because stuff got moved around. Here's a summary of the technical changes for you:
- Extract `ExploreFocusSplit` from `App`
  - There's new routing logic here
- Extract `app-layout.tsx` from `App`
  - Both `App` and `ExploreFocusSplit` need these styles
- Extract `HeaderCountrySelect` from `Header`
  - There's new routing logic here
- Extract `FocusVariantHeaderControls` from `VariantHeader`
  - Only `FocusPage` needs these controls
  - A few other pages now use `VariantHeader` (without controls)
- Rename `SamplePage` to `GlobalSamplePage`
- Create `DeepFocusPage`
  - Contains basic routing and styles


chaoran-chen:
Quick comment about the international part: I would suggest adding "global" and the continents as options to the country selection in the future.

tehwalris:
> Quick comment about the international part: I would suggest adding "global" and the continents as options to the country selection in the future.

At the moment that's not really practical since some API endpoints require country (eg. `/plot/variant/age-distribution`). If `country` is consistently optional everywhere in the API, then I think allowing "global" in the dropdown is a good design.

chaoran-chen:
Could you please create a redirect from our old `/variant` to `/explore/Switzerland`? It's just for now because some people might already have bookmarked the old link but we can remove the redirect in a few weeks.

chaoran-chen:
Nothing seems to break, and the URL scheme is not beautiful but I agree that it's easy to implement! So 👍 from my side

spacephoton:
Having trouble merging changes. Could one of you do it? Seems like most of the conflicts are between both of your changes, so I am not sure how to fix.

@chaoran-chen @tehwalris

chaoran-chen:
Thanks, everyone!

PR #37, opened 2021-03-05
Add gene information tooltip

### LINKED ISSUE #30: Show general information about genes
_Taking "N:P80R" as an example, do we know something about the N or even about N:80?_

When the user clicks on "N", a short description (and maybe some references) about the N-gene should be shown in a tooltip. Later, we will integrate more detailed information about certain regions within a gene.

chaoran-chen:
Close #30 

Many thanks to @Julia91457 for sharing her knowledge about the proteins!!

### DIFF
--- a/src/components/MutationList.tsx
+++ b/src/components/MutationList.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { MutationName } from './MutationName';
+
+interface Props {
+  mutations: string[];
+}
+
+export const MutationList = ({ mutations }: Props) => {
+  return (
+    <>
+      {mutations
+        .map<React.ReactNode>(m => <MutationName key={m} mutation={m} />)
+        .reduce((prev, curr) => [prev, ', ', curr])}
+    </>
+  );
+};
--- a/src/components/MutationName.tsx
+++ b/src/components/MutationName.tsx
@@ -1,0 +1,186 @@
+import React from 'react';
+import styled from 'styled-components';
+import { OverlayTrigger, Popover } from 'react-bootstrap';
+
+interface Props {
+  mutation: string;
+}
+
+const headerMap: { [index: string]: string } = {
+  ORF1a: 'ORF1ab polyprotein',
+  ORF1b: 'ORF1ab polyprotein',
+  S: 'Spike glycoprotein',
+  E: 'Envelope protein',
+  M: 'Membrane glycoprotein',
+  N: 'Nucleocapsid phosphoprotein',
+};
+
+const textMap: { [index: string]: string } = {
+  S:
+    'important for binding to host cell, trimeric, can be cleaved into S1 part (containing RBD) and S2 part; ' +
+    'S1 = involved in attachment of virions by interaction with human ACE2; S2: fusion protein; immunogenic',
+  E:
+    'relevant for assembly and release of virions, 75 aa, can assemble into host membrane to form ' +
+    'protein-lipid pores (viroporin)',
+  M: '222 aa, relevant for RNA packaging, most abundant viral protein, contains transmembrane domains',
+  N: 'relevant for packing of RNA into ribonucleocapsid, known to be immunogenic',
+};
+
+const ORF1abProteins = [
+  {
+    range: [1, 180],
+    text:
+      "N-terminal product of viral replicase, mediates RNA replication and processing, suppresses hosts' immune functions",
+    name: 'Nsp1',
+  },
+  {
+    range: [181, 818],
+    text: 'replicase product essential for proofreading, modulation of host cell signalling pathways',
+    name: 'Nsp2',
+  },
+  {
+    range: [819, 2763],
+    text: 'papain-like protease, separates polyprotein into distinct proteins',
+    name: 'Nsp3',
+  },
+  {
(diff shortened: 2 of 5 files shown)


src/components/MutationName.tsx
```diff
@@ -0,0 +1,186 @@
...
+      <Popover.Title as='h3'>{information.header}</Popover.Title>
+      <Popover.Content>
+        <div style={{ width: '600px' }}>{information.text}</div>
+      </Popover.Content>
+    </Popover>
+  );
+
+  return (
+    <>
+      {/*TODO It does not seem right to overwrite the CSS here. But how else?*/}
```
chaoran-chen, line 173: This looks like a bad idea.. Do you know how to implement it properly?

src/components/MutationList.tsx
```diff
@@ -0,0 +1,16 @@
...
+interface Props {
+  mutations: string[];
+}
+
+export const MutationList = ({ mutations }: Props) => {
+  return (
+    <>
+      {mutations
+        .map<React.ReactNode>(m => <MutationName key={m} mutation={m} />)
+        .reduce((prev, curr) => [prev, ', ', curr])}
```
tehwalris, line 13: This reduce is not making a flat list (try `['a', 'b', 'c'].reduce((a, c) => [a, ', ', c])` in your console for an example). It still renders because arrays of `ReactNode` are also `ReactNode`.

I suggest `.flatMap((v, i) => (i === 0 ? [v] : [', ', v]))` or something. That way you can also remove the explicit type parameter.

src/components/MutationName.tsx
```diff
@@ -0,0 +1,186 @@
...
+const Mut = styled.span`
+  &:hover {
+    background: #eaeaea;
+    border-radius: 15px;
+    padding: 5px;
+    cursor: pointer;
+  }
+`;
+
+const getInformation = (gene: string, position: number): { header: string; text: JSX.Element } => {
```
tehwalris, line 119: If this returned `{header: string, text?: string}` the separation of logic and rendering would be cleaner. You could handle the "no description" text and styles in the main part of the component.

src/components/MutationName.tsx
```diff
@@ -0,0 +1,186 @@
...
+  },
+  {
+    range: [6799, 7096],
+    text: "2'-O-ribose methyltransferase",
+    name: 'Nsp16',
+  },
+];
+
+const Mut = styled.span`
+  &:hover {
```
tehwalris, line 111: When you hover on the mutations everything jumps around and that's really annoying. It would be better to apply the padding and border-radius outside of `:hover`.

You could try `  border-radius: 5px; padding: 3px 5px; margin-right: -5px;` to line it up nicely with the text.

src/components/MutationName.tsx
```diff
@@ -0,0 +1,186 @@
...
+    <>
+      {/*TODO It does not seem right to overwrite the CSS here. But how else?*/}
+      <style type='text/css'>
+        {`
+          .popover {
+            max-width: none;
+          } 
+        `}
+      </style>
+      <OverlayTrigger trigger='click' overlay={popover} rootClose={true} transition={false}>
```
tehwalris, line 181: Some of the larger popups get slightly cut off. Adding `placement='bottom'` should be good enough, since we have plenty of space below.

cut-off-popup

src/components/MutationName.tsx
tehwalris: In this specific case add `style={{ maxWidth: '600px' }}` to `Popover` (and you can remove your inner `div`).

In general in most of react-bootstrap you can pass a styled component to `as`. In most other libraries you can convert the styled component to a `className` and pass it that way. In this case `Popover` is mean doesn't support either of those (and it's rendered out-of-tree, so it won't immediately work to wrap it and then used scoped selectors).

tehwalris: BTW The way this was (`<style type='text/css'>`) the styles would apply to the whole application, not just this component.

src/components/NewVariantTable.tsx
```diff
@@ -46,7 +47,9 @@ export const NewVariantTable = ({ country, year, week, onVariantSelect }: Props)
             <tbody>
               {data.map(d => (
                 <tr key={d.variant.mutations.join(',')}>
-                  <td style={{ maxWidth: '400px', lineBreak: 'auto' }}>{d.variant.mutations.join(', ')}</td>
+                  <td style={{ maxWidth: '400px', lineBreak: 'auto' }}>
+                    <MutationList mutations={d.variant.mutations} />
```
tehwalris, line 51: My browser is not happy rendering all of these React components and DOM elements here in this huge table. Can we keep `d.variant.mutations.join(', ')` within this table?

tehwalris: To see the issues: load all of the B1.1.7 samples, resize the width of your page and then try scrolling.

tehwalris (changes requested):
The feature works nicely. The code is good overall. I added a few minor change requests and comments.

chaoran-chen:
Thanks for the review, @tehwalris. Because the PR is now merged, I created a new issue (#42) to address your comments.

Issue #10, opened 2021-02-17
International comparison plot: logarithmic scale view


chaoran-chen:
The user should be able to switch the scale of the international comparison plot to logarithmic. Data points with y=0 should be omitted (since log(0) is undefined).

For B.1.1.7, it would look like this:

![image](https://user-images.githubusercontent.com/18666552/108223582-41bed780-713a-11eb-9fef-9ba491aa2dc4.png)

(Source: https://ibz-shiny.ethz.ch/covidDashboard/variant-plot/index.html)

PR #45, opened 2021-03-06
Fill missing data with zeros

### LINKED ISSUE #15: Nothing = Zero
For all the plots we currently have, we can safely assume that the response contains the full dataset and weeks/ages/etc., that are not mentioned, occurred zero times.

For example...

image

Here, we know that between the weeks 23 and 26, the variant was never sequenced. Further, making the assumption that we did perform sequencing through the whole time, we can also set the proportion to 0%.

tehwalris:
This closes #15. It fills data in all of our plots (time, demographics and international). The plots where x is time are filled between the first and last week that is present in each series (so countries in international are filled independently). The age plot is filled according to our fixed buckets ("0-9", "10-19", etc). The log plot (international) now fillters out zero values (since they exist now) as originally requested in #10.

I added lodash, dayjs (a date library) and some helper functions for working with year-week strings - you might find those useful in the future. 

Note I removed confidence intervals from the state, since we are not showing them anywhere, and zero-filled values don't really have a defined confidence interval. We can add them back and make them optional later if we ever need them.

### DIFF
--- a/.github/workflows/react.yml
+++ b/.github/workflows/react.yml
@@ -28,3 +28,4 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm run check-types
+      - run: npm run test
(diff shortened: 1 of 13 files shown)


spacephoton:
Surprise it was so complex w/ the dates, but seems to work great!! Awesome!

PR #46, opened 2021-03-06
Compact layout in focus panel

### LINKED ISSUE #38: Compact layout in focus panel
Currently all the charts in the "focus" panel are stacked vertically. That means you almost always have to scroll, and some charts are way too wide (eg. age distribution). We could put some of these plots side by side on a bigger screen. When we do that, we should consider adding borders, backgrounds or shadows to keep the layout visually clear.

tehwalris:
This closes #38. The idea is that every section on the focus page has a minimum width. As many things as possible should be displayed side by side (in a row), as long as their minimum width is respected. If there is extra space in a row, the sections should be scaled to fill it.

I first thought about how to do this with CSS and couln't come up with a good plan. I then implemented the layout in JS, which works well (`feature/compact-layout`). Then I realized that `flex-wrap` would probably make this possible with just CSS, so I deleted all my JS code and tried to do it that way (`feature/compact-layout-css`).

The CSS version would be good, since it uses less code. Sadly I think it's impossible to get it to work with just CSS *in our application*. The basic idea with `flex-wrap` works and does exactly what it should. The problem is that we have components that measure their size in JS (the map, but also plotly internally). When the page loads, they measure their size and render their content at that size. If you now resize your window, the components will still be taking up their originally measured size, so flex will not adjust its wrapping, since it thinks that there is no space. I also saw some weird effects with components smoothly resizing (like animating to a different size by themselves for no reason) when the page loaded (without resizing the window) - that's probably also an interaction of flex and measurement.

Condsidering those issues I suggest keeping the JS version. From my experience building a visualization tool that had a layout with lots of plots, doing the top level layout in CSS eventually becomes impossible (or at least difficult and unstable) because of such measurement problems.

The JS version consists of a `PackedGrid` component which you put `GridCell` children into. `PackedGrid` collects all the layout information (`minWidth`, etc) from the props of those children and calls `placeGridCells`. `placeGridCells` calculates which rows elements should go on (more or less as described in the first paragraph) and calculates their widths. `PackedGrid` then renders each child wrapped in a `div` in the calculated row and with the calculated width.

With the JS version as described above, elements would remount when they moved across rows (React considers elements that move between divs to be new elements). I used `react-reverse-portal` in `PackedGrid` to avoid this remounting.


spacephoton:
Seems functionally good, but I dont' like the different sizes for the top row:
image

Maybe we can do like this:
[Sequences over time][Demographics] (half-half)
[Geography]

What do you think?
@tehwalris @chaoran-chen

spacephoton:
and isn't sequenec over time the most important? how does it compare to international comparison in terms of importance?

chaoran-chen:
Having three plots in a row might be a little too much but I find it difficult to judge it right now and would suggest that we try out the current version.

The international comparison plot is in general quite useful but I think that we will move it so somewhere else entirely in the future.

spacephoton:
I like the changes to the grid, and suggest we keep the three graphs line by line instead of putting them on the same line and merge. Other than that everything seems great.

What do you think? @chaoran-chen @tehwalris

tehwalris:
I've changed the settings to split less. On screens below ~1950 CSS pixels width (most screens), nothing will be side by side. On larger screens "Sequences over time" and "Demographics" will be side by side, dividing the width in a 2:1 ratio. "Sequences over time" will be at least 800px wide when that happens. It should look nicer once the demographics plot is updated to match. Here's how that looks:

new-layout

spacephoton:
Seems good. I don't like the aesthetics too much,b ut might look better once im done matching the demographics graph.
Anyways, it's an edge case so can merge.

PR #47, opened 2021-03-07
Feature/newtimeplot

### LINKED ISSUE #43: Simplified sequence over time with recharts
- [x] MVP:
image
- [ ] Eventually, be able to click and drag to select multiple times.
image

spacephoton:
CleanShot 2021-03-07 at 14 16 01
MVP done, let me know if any bugs. 

For now, have to hover on the bar itself to select it. This means that bars with 0 value will not be hoverable.

The dates could be aligned better, but might not be worth the work.

Aesthetics can be improved, but that's mostly moving things around. For now I think it is okay.

There is a click handler for when we click a bar, which might be useful.


chaoran-chen:
Hi. This looks already really nice! I have just one request: Could you please add some fixed x-axis and y-axis ticks and titles so that the user can also understand what's going on without hovering and, for example, take a useful screenshot.

chaoran-chen:
To make it easier to understand the overall prevalence at the first glance, what do you think about fixing the y-axis' range to [0, 1]?

chaoran-chen:
Oh, and one more thing. Could you please add the week number as well? In many other plots/reports, we also use the week number so it would be nice to show both the date and the week number.

spacephoton:
> To make it easier to understand the overall prevalence at the first glance, what do you think about fixing the y-axis' range to [0, 1]?

Wouldn't this mean that the datapoints are very small for many of the selected mutations? since we are talking about percentages compared to other samples

TKGZ:
For example:

spacephoton:
image

spacephoton:
Not too happy with the alignment, but other than that is this what you had in mind? @chaoran-chen 

CleanShot 2021-03-07 at 15 49 37

spacephoton:
Actually, I don't think we should put the ticks on the bottom. It doesn't convey much useful information. This way we can focus on the final date, and still have an idea of the general trends. If we want to know exact date we can scroll mouse.

chaoran-chen:
Yes, the y-axis looks great. And I do believe that having x-axis ticks is also very important. The user should be able to take screenshots of the plots and be able to include them in presentations or reports, and for that, having ticks is inevitable. Also, in general, I find it much harder (and slower) to grasp what a plot shows if there are no clear ticks. And I really would not be surprised if some user will not even get the idea to use the mouse to hover over the bars.

So, I think that it would be great if we could have static month(+year) ticks on the x-axis.

chaoran-chen:
Few minor suggestions: "Proportion of the variant on the week of...", a little smaller font (the below screenshot uses 1.2rem), and a bigger padding-bottom.

image

spacephoton:
> Few minor suggestions: "Proportion of the variant on the week of...", a little smaller font (the below screenshot uses 1.2rem), and a bigger padding-bottom.

padding between title and graph?

chaoran-chen:
Yes, i.e. between "Proportion of the variant..." and the plot.

spacephoton:
> Yes, the y-axis looks great. And I do believe that having x-axis ticks is also very important. The user should be able to take screenshots of the plots and be able to include them in presentations or reports, and for that, having ticks is inevitable. Also, in general, I find it much harder (and slower) to grasp what a plot shows if there are no clear ticks. And I really would not be surprised if some user will not even get the idea to use the mouse to hover over the bars.
> 
> So, I think that it would be great if we could have static month(+year) ticks on the x-axis.

Ok, so now I added back the ticks.
CleanShot 2021-03-07 at 17 48 02

The year gets added when there is enough space. 
image

image

chaoran-chen:
Thanks!! From my side, it's ready to merge!

chaoran-chen:
I will just go ahead and merge it because I like it a lot :)

chaoran-chen (approved):

spacephoton:
yay! :D

src/widgets/VariantTimeDistributionPlot.tsx
```diff
@@ -1,17 +1,34 @@
...
+import ReactTooltip from 'react-tooltip';
+
+import { BarChart, XAxis, YAxis, Bar, Cell, ResponsiveContainer, CartesianGrid } from 'recharts';
+import styled from 'styled-components';
+
+import { BiHelpCircle } from 'react-icons/bi';
 import { fillWeeklyApiData } from '../helpers/fill-missing';
 import { EntryWithoutCI, removeCIFromEntry } from '../helpers/confidence-interval';
 
+const CHART_HEIGHT = 290;
```
tehwalris, line 17: We can't have a fixed height, especially in the embed. I've [made the plot fill the available space](https://github.com/cevo-public/cov-spectrum-website/commit/3b69001e30a3d4a18bfead2e2086369b182303e8).

src/widgets/VariantTimeDistributionPlot.tsx
```diff
@@ -36,59 +53,288 @@ export const VariantTimeDistributionPlot = ({ country, mutations, matchPercentag
     };
   }, [country, mutations, matchPercentage]);
 
+  const processedData: TimeEntry[] | undefined = distribution
```
tehwalris, line 56: `distribution?.map(...)`

src/widgets/VariantTimeDistributionPlot.tsx
```diff
@@ -36,59 +53,288 @@ export const VariantTimeDistributionPlot = ({ country, mutations, matchPercentag
...
+  quantity: number;
+};
+
+export type TimeGraphProps = {
+  data: TimeEntry[];
+  height?: number;
+  onClickHandler?: OnClickHandler;
+};
+
+export const TimeGraph = React.memo(
```
tehwalris, line 220: `TimeGraph`, `Metric` etc should be moved to their own file(s) when we re-use them for the "demographics" plot.

src/widgets/VariantTimeDistributionPlot.tsx
```diff
@@ -36,59 +53,288 @@ export const VariantTimeDistributionPlot = ({ country, mutations, matchPercentag
...
+  ({ data, onClickHandler, height = CHART_HEIGHT }: TimeGraphProps): JSX.Element => {
+    const [activeIndex, setActiveIndex] = useState<number>(data.length - 1);
+    const [ready, setReady] = useState(false);
+    const [currentData, setCurrentData] = useState<TimeEntry>(data[data.length - 1]);
```
tehwalris, line 224: I think if `data` is empty this could crash (eg. when accessing `currentData.yearWeek`).

tehwalris (commented):
Very nice result!

src/widgets/VariantTimeDistributionPlot.tsx
tehwalris: Ah, this won't actually crash because you have `&& current Data ?` below. This line should still be changed to `useState<TimeEntry | undefined>` so that if someone uses `currentData` without checking there will be a TS error.

tehwalris: If `data` is empty you do get some errors from recharts though. `Received NaN for the "y1" attribute. If this is expected, cast the value to a string.` That's fine for now, since they don't crash the application.

You can search for `N:R203K, N:G204R, S:N501Y, S:A123B, S:N507Y, N:R555Y` and with 100% match percentage to test with empty `data`.

src/widgets/VariantTimeDistributionPlot.tsx
spacephoton: Perfect, thank you! :D

src/widgets/VariantTimeDistributionPlot.tsx
spacephoton: Good point

spacephoton:
- [x] remove empty space on the left

spacephoton:
- [x] make sequence tick when > 20 bars to display
image

spacephoton:
- [x] add percent on left bar

spacephoton:
CleanShot 2021-03-13 at 17 05 11
Won't keep the proportion title shifted, as it looks worse like this.  But rest should is done.

PR #53, opened 2021-03-12
Random-samples-only filter

### LINKED ISSUE #31: Random samples-only filter
In the top bar, next to the country selection, another filter should be added. The user should be able to choose between "all samples" and "surveillance". The "surveillance"-option should be only available for Switzerland (and otherwise grayed out?). If surveillance is selected, the international comparison plot should be hidden.

The default (for Switzerland) should show all samples.

This issue depends on https://github.com/cevo-public/cov-spectrum-server/issues/10. [The API](https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md) was updated: all appropriate endpoints now have an additional, optional query param called `dataType`. If nothing is provided, all samples will be used, and if `SURVEILLANCE` is passed, the server will only use selected samples.

tehwalris:
This closes #31. It adds a "Sampling strategy" dropdown in the header next to the country select. It is hidden on all of the screens where it is not applicable. It is shown but disabled when the country is not Switzerland.

In general the setting is not saved in the URL or in local storage. We could put it in local storage in the future, but putting it in the URL will be hard. Note that the setting is saved in the URL of embeded plots.

Since the international comparison endpoints don't support only showing surveillance samples, but that component is still in the focus panel, I added a warning message if the sampling strategy is not set to "All samples".

@chaoran-chen I'm not sure if the server is handling the filter correctly for `/plot/variant/time-zip-code-distribution`. I checked that the query parameter is added in the request, but I can't see a visual change on the map.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,9 +12,9 @@ import { ExploreFocusSplit } from './pages/ExploreFocusSplit';
 import { LoginPage } from './pages/LoginPage';
 import { GlobalSamplePage } from './pages/GlobalSamplePage';
 import { useSamplingStrategy } from './components/HeaderSamplingStrategySelect';
+import { SamplingStrategy } from './services/api';
 
 export const App = () => {
-  // TODO
   const samplingStrategyProps = useSamplingStrategy();
 
   return (
@@ -25,7 +25,7 @@ export const App = () => {
 
       <Switch>
         <Route exact path='/'>
-          <Redirect to='/explore/Switzerland' />
+          <Redirect to={`/explore/Switzerland/${SamplingStrategy.AllSamples}`} />
         </Route>
         <Route path='/variant'>
           {/* This is so that we don't break old bookmarked links */}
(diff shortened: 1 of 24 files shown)


chaoran-chen:
Hi @tehwalris! The time-zip-code-distribution will not show changes right now because we only have zip code information for our data.

What do you think about `/explore/{country}` for all and `/explore/{country}-{dataType}` if a dataType is selected? I assume that we will have even more types of data in the future (for example, "animals") so that having direct links to a certain category of data would be really great.

tehwalris:
> Hi @tehwalris! The time-zip-code-distribution will not show changes right now because we only have zip code information for our data.
> 
> What do you think about `/explore/{country}` for all and `/explore/{country}-{dataType}` if a dataType is selected? I assume that we will have even more types of data in the future (for example, "animals") so that having direct links to a certain category of data would be really great.

I think we could put it in the URL (maybe as `/explore/:country/:samplingStrategy/` though). I thought it would be quite a lot of work, but I looked at it again now and it might be fine. I'll set this PR to draft and try it out, then I'll set the PR back to normal once it's done.

chaoran-chen:
Great, thank you!

tehwalris:
I changed everything so that the sampling strategy is in the URL (`/explore/:country/:samplingStrategy`). Overall I'm happy with how it turned out.

I use string operations to modify the URL for redirects or changing values (eg. country). That isn't super nice, but it's isolated in one file and I've tested it well enough.

spacephoton:
Seems great to me!

PR #57, opened 2021-03-14
Add chen2021Fitness model

### LINKED ISSUE #50: Fitness advantage estimation model
Even though already public for a while, I still need to create a PR. What's missing is a description of the model and some code cleaning.

chaoran-chen:
This closes #50.


src/pages/FocusPage.tsx
```diff
@@ -42,6 +43,10 @@ export const FocusPage = (props: Props) => {
           <Switzerland {...plotProps} />
         </NamedSection>
       )}
+      <NamedSection title='Model'>
+        {/*TODO Should we make height optional?*/}
+        <Chen2021FitnessWidget.ShareableComponent {...plotProps} height={-1} />
```
chaoran-chen: The height parameter does not fit. Do you have a better suggestion, or will it be part of the planned layout restructuring anyway?

src/models/chen2021Fitness/Chen2021FitnessWidget.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+const ResultsPropsSchema = zod.object({
+  request: Chen2021FitnessRequestSchema,
+});
+type ResultsProps = zod.infer<typeof ResultsPropsSchema>;
+
+type Chen2021FitnessRequest = zod.infer<typeof Chen2021FitnessRequestSchema>;
+type Chen2021FitnessResponse = zod.infer<typeof Chen2021FitnessResponseSchema>;
+
+const getData = async (
```
chaoran-chen: Would you prefer moving it to `api.ts` or should we leave it here?

spacephoton: I dont know

tehwalris: Let's keep it separate from other API functions for now.

src/models/chen2021Fitness/Chen2021FitnessWidget.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+const ContainerPropsSchema = SampleSelectorSchema;
+type ContainerProps = zod.infer<typeof SampleSelectorSchema>;
+
+const ResultsPropsSchema = zod.object({
+  request: Chen2021FitnessRequestSchema,
+});
+type ResultsProps = zod.infer<typeof ResultsPropsSchema>;
+
+type Chen2021FitnessRequest = zod.infer<typeof Chen2021FitnessRequestSchema>;
```
tehwalris: `ResultsProps` and `Chen2021FitnessRequest` should not use Zod. It is important for API _responses_ or props loaded _from a URL_ (data we get from the server or user), since invalid data will cause crashes and weird bugs, even though the TS types say they shouldn't happen. For requests it just complicates things and makes TS print less nice errors.

Note that it's still required to use Zod for `ContainerProps`, since that is used to check the props when they get loaded from the URL.

src/models/chen2021Fitness/Chen2021FitnessWidget.tsx
```diff
@@ -0,0 +1,381 @@
...
+type Chen2021FitnessResponse = zod.infer<typeof Chen2021FitnessResponseSchema>;
+
+const getData = async (
+  params: Chen2021FitnessRequest,
+  signal: AbortSignal
+): Promise<Chen2021FitnessResponse | null> => {
+  const mutationsString = params.mutations.join(',');
+  const url =
```
tehwalris: [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) is designed exactly for building URLs like this. You could use `new URLSearchParams(params)` to set most values and then `.set(...)` and `.delete(...)` to handle the few special fields.

src/models/chen2021Fitness/Chen2021FitnessWidget.tsx
```diff
@@ -0,0 +1,381 @@
...
+    `&reproductionNumberWildtype=${params.reproductionNumberWildtype}` +
+    `&plotStartDate=${params.plotStartDate.toISOString().substring(0, 10)}` +
+    `&plotEndDate=${params.plotEndDate.toISOString().substring(0, 10)}` +
+    `&initialWildtypeCases=${params.initialWildtypeCases}` +
+    `&initialVariantCases=${params.initialVariantCases}`;
+  const response = await get(url, signal);
+  if (response.status !== 200) {
+    // The computation might fail, for example, if some values go out-of-bound. The issue shall be addressed with the
+    // introduction of a better error handling and reporting on the server side.
+    return null;
```
tehwalris: If theres no particular reason to use `null`, it's a bit nicer to use `undefined` so that you can write `| undefined` instead of `| null | undefined` in types.

src/models/chen2021Fitness/Chen2021FitnessWidget.tsx
```diff
@@ -0,0 +1,381 @@
...
+const Wrapper = styled.div`
+  background-color: #ffe0b6;
+  padding: 15px;
+`;
+
+const SectionHeader = styled.h5`
+  margin-top: 20px;
+`;
+
+const Chen2021FitnessContainer = ({
```
tehwalris: It's very hard to keep an overview of React code when multiple components are in one file. In some cases there are good reasons to put everything in one file, but in this case I think that the components can easily be put in separate files.

src/models/chen2021Fitness/Chen2021FitnessWidget.tsx
```diff
@@ -0,0 +1,381 @@
...
+    paramData.reproductionNumberWildtype.toString()
+  );
+  const [formInitialVariantCases, setFormInitialVariantCases] = useState(
+    paramData.initialVariantCases.toString()
+  );
+  const [formInitialWildtypeCases, setFormInitialWildtypeCases] = useState(
+    paramData.initialWildtypeCases.toString()
+  );
+
+  useEffect(() => {
```
tehwalris: I see how `useEffect` is useful in this specific case, so it's fine here. However in general `useMemo` is the right hook for this sort of thing.

src/models/chen2021Fitness/Chen2021FitnessWidget.tsx
```diff
@@ -0,0 +1,381 @@
...
+      isSubscribed = false;
+      controller.abort();
+    };
+  }, [request]);
+
+  if (!modelData) {
+    return <>A fitness advantage cannot be estimated for this variant.</>;
+  }
+
+  const filteredDaily: zod.infer<typeof Chen2021FitnessResponseSchema.shape.daily> = {
```
tehwalris: `Chen2021FitnessResponse["daily"]` is a simpler way to write that type.

tehwalris (dismissed):
Great that this being merged to develop. You should not use Zod in some places, as I described. The other comments are less important.

chaoran-chen:
Thanks for the reviews!

PR #60, opened 2021-03-18
Nicer "Known variants"

### LINKED ISSUE #59: Better "Known variants" list
With some kind of basic stats or plots to help decide which variants to select

tehwalris:
new-known-variants

This PR replaces the "Known variants" table with little cards. The idea is to give the user a quick overview over the behavior of the known variants.

The plots show the latest 8 weeks of data that we have, with the same x scale across all plots. Every plots y-axis is scaled to fill the available vertical space, so that you can see how each variant is developing. There's also a small indicator that shows the proportion of all sequences, so you can understand how the variants compare to eachother.

Note that I've reordered `known-variants.json` so that it's sorted by the current prevalence in Switzerland.


chaoran-chen:
Looks beautiful and is very space-efficient!

spacephoton:
@tehwalris This design looks incredible!! Super simple, super elegant, I think it looks much better than my graphs. Awesome job! Maybe a little thing, when there are few or no cases, it's not clear that we have selected that variant. But other than that, seems to work great. Same for interesting variants. But can fix later. :)

PR #61, opened 2021-03-19
New demographics chart

### LINKED ISSUE #51: New style for age distribution plot
As suggested by @TKGZ, we would like to use the design of the time distribution plot for the age distribution plot as well.

spacephoton:
#51 done
- Extracted and separated chart components. Not 100% separated, but we can do that if we need to reuse this type of chart later. Updated demo chart to new style. 
- Added loading spinner when changing variant selection/loading.


chaoran-chen (changes requested):
Hi @TKGZ! Fantastic, that's so much better than the old plot!!

Only one thing: the explanation tooltips should be adapted:

image
image

Maybe
"Proportion relative to all samples collected from this age group"
"Number of samples of the variant collected from this age group" ?

chaoran-chen (approved):
From the "user test perspective" - 👍

spacephoton:
i'll go ahead and merge since not many changes and i would like to reuse this code for the other chart.

tehwalris (commented):
The separation seems good. Thanks for adding the spinners! 

BTW You could have left this PR open and based the new one off this branch, then changed the merge target to `develop` later. This PR would still have to be merged first though. Anyway the way you did it is fine too.

PR #66, opened 2021-03-20
Feature/sequencingchart

### LINKED ISSUE #44: Sequencing intensity plot
In the explore area, we would like to show a plot presenting the sequencing intensity of the selected country.

The corresponding API is at https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md#sequencing-intensity-through-time

A suggestion:

image

On hover, a tooltip should show the absolute number of cases, the absolute number of sequenced samples, and the relative number as percentage.

spacephoton:


spacephoton:
Ugly MVP but I'd like feedback first

image

What colors should I use?
Suggestions for fixing lack of space? Should I use tooltips instead?

chaoran-chen:
Hi @TKGZ. Could we try out monthly bars? I would be very interested to see how that looks like.

spacephoton:
> Hi @TKGZ. Could we try out monthly bars? I would be very interested to see how that looks like.

might have to have different api since the weeks cross months right?

Might indeed make more sense to have it, or maybe we could have a log scale?

chaoran-chen:
Yes, of course! I changed the endpoint to return daily data.

I actually like the message that we are sequencing only sooo few samples relative to the case numbers. Only very few countries - Australia, New Zealand, Iceland, Denmark, and Singapore - are sequencing >= 10% of the cases.

spacephoton:
Confirmed cases include sequenced cases right? So non-sequenced are confirmed - sequenced?

chaoran-chen:
Yes, exactly. But because the data come from different sources and some countries don't report new confirmed cases on weekends, it is possible that on some days, the number of sequenced cases is larger than the number of confirmed cases.

src/charts/TimeIntensityChart.tsx
```diff
@@ -0,0 +1,165 @@
...
+};
+
+export const TimeIntensityChart = React.memo(
+  ({ data, onClickHandler }: Props): JSX.Element => {
+    const [activeIndex, setActiveIndex] = useState<number>(data.length - 1);
+    const [ready, setReady] = useState(false);
+    const [currentData, setCurrentData] = useState<TimeIntensityEntry>(data[data.length - 1]);
+
+    useEffect(() => {
+      setReady(true);
```
tehwalris: What is the point of this?

src/widgets/SequencingIntensityPlot.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Widget } from './Widget';
+import * as zod from 'zod';
+import { ZodQueryEncoder } from '../helpers/query-encoder';
+import TimeIntensityChart from '../charts/TimeIntensityChart';
+import Loader from '../components/Loader';
+
+const PropsSchema = CountrySelectorSchema;
+type Props = zod.infer<typeof PropsSchema>;
+
+const processData = (data: SequencingIntensityEntry[]): any =>
```
tehwalris: Don't forget to remove the `any` - it will lead to bugs.

tehwalris (commented):
I can't test this at the moment, since it's broken because the data is now per day. It looks great on the screenshot though. I like the linear scale for the same reason as Chaoran. I guess for days where confirmed < sequenced the bar should be the height of sequenced.

chaoran-chen:
..alternatively, we could of course present the same plot as for time and age distribution again and only show the relative numbers (with the absolute numbers on the right/as tooltip). If the bar for the number of sequenced samples is just never visible, that might be better.. What do you prefer, @TKGZ?

spacephoton:
Actually I think your previous point about linearity works. Will try to make it by month and linear and show u

spacephoton:
Color is ugly, but we can fix later :)

src/components/NewVariantTable.tsx
```diff
@@ -47,8 +47,8 @@ export const NewVariantTable = ({ country, onVariantSelect }: Props) => {
   const variants = useMemo(() => {
     const uniqueVisibleMutations = new Set();
     return data?.variants
-      .map(v => {
-        const visibleMutations = v.mutations.filter(m => {
+      .map((v: any) => {
```
tehwalris, line 50: Why did you add `any` here? I've [reverted it](https://github.com/cevo-public/cov-spectrum-website/commit/171aff70d2e31892ce765d0e0fa0bf113df096b4) and it looks like the correct type gets inferred.

PR #72, opened 2021-03-25
Improve structure of focus pages

tehwalris:
This PR makes the major structural parts of the style changes that I presented at the last meeting:

  - Add cards around the things on the focus page
  - Move most of "Fitness advantange estimation" and "International comparison" to deep focus pages

Note that this PR on purpose does not make (most) cosmetic changes to the style (colors, shadows, etc). I'll do those in a future PR (see issue https://github.com/cevo-public/cov-spectrum-website/issues/73).

Please tell me if you notice any general issues or bugs with the restructured UI. I'm not really asking for technical feedback, since structural changes are tough to read in a PR.

### DIFF
--- a/src/components/FocusVariantHeaderControls.tsx
+++ b/src/components/FocusVariantHeaderControls.tsx
@@ -29,7 +29,7 @@ export const FocusVariantHeaderControls = ({
           samplingStrategy: toLiteralSamplingStrategy(samplingStrategy),
         })
       }
-      variant='outline-dark'
+      variant='secondary'
       size='sm'
       className='mr-2'
     >
@@ -56,7 +56,7 @@ export const FocusVariantHeaderControls = ({
       )}
       <LazySampleButton
         query={{ variantSelector: { variant, matchPercentage }, country, samplingStrategy }}
-        variant='outline-dark'
+        variant='secondary'
         size='sm'
       >
         Show samples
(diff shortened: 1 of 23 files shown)


chaoran-chen (changes requested):
Could we give the age distribution plot a larger min-width? On my screen, the x-axis labels are not readable at all:

image

Otherwise, it works great!

spacephoton:
Except for the sizing issue on large screens, I think it's ready for merge.
image

spacephoton:
Shall we just put these on two rows instead of one? 
There isn't much advantage in keeping on same row

image

tehwalris:
I limited the layout to at most 2 columns, even on large screens. I also made the sequences over time and demographics plots have an equal width, as @TKGZ once suggested. Here's how that looks on my screen:

full-screen

I increased the minimum width required for the demographics plot to be split, but that wasn't enough to stop the labels from overlapping. I tried hiding the labels when there isn't enough space, but that looks really bad. In the end I chose to show a shorter form of labels when the plot is narrow. Here they are at the smallest width which the demographics plot can have in a two-column layout:

smallest-two-column-width

The smallest width at which full labels are still used looks like this:

smallest-full-label-width

tehwalris:
I forgot to push commit 69781fb8864beee40c35e7c8dbc54c2ffc74a7d1 before merging this PR. I merged it to `develop` separately.

spacephoton:
Awesome seems great!

PR #78, opened 2021-04-06
Use new samples endpoint

### LINKED ISSUE #62: Switch to the general samples endpoint
Following @tehwalris' idea, a new, general endpoint to get sample information was created. It should be used to calculate all the variant-specific plots and render the `/plot/variant/*` endpoints obsolete.

Documentation: https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md#sample-new

### LINKED ISSUE #65: When no sample can be found..
When no sample can be found, the focus page should show a clear note instead of a set of empty spaces.

tehwalris:
This PR closes #62. It switches to the new samples endpoint wherever possible. Everything seems to work exactly as before, but please double check for new bugs. Note that this PR makes **breaking changes to embed URLs** (see the "Design" > "Embeds" section).

## Where is the new endpoint used?
The new samples endpoint is used for:
- The"Sequences over time" plot
- The "Demographics" plot
- The Swiss map
- "International comparison" (including table)
- The small plots on the "Known variants" buttons

The new samples endpoint is _not_ used for:
- "Interesting variants"
- "Fitness advantage estimation"
- "Show samples"
- "Show on Nextclade"

I think it's fundamentally not possible to use the new endpoint for any of the above. Tell me if you see a way.

### Data loading
The new API endpoint is called using `getNewSamples` from `api.ts`. `getNewSamples` returns a `SampleSet`, which is a wrapper around the list of samples from the REST API. `SampleSet` contains methods for the most common operations on this data.

`getNewSamples` calls happen in `ExploreFocusSplit`, because the data is needed by most components below. Specifically, 4 sample sets are loaded here:

- `variantSampleSet` - filtered by mutation and country
- `wholeSampleSet` - filtered by country
- `variantInternationalSampleSet` - filtered by mutation
- `wholeInternationalSampleSet` - not filtered

Most components will take `variantSampleSet` and `wholeSampleSet`, then use `SampleSet.proportionBy(Week|Field)` and `fillFrom(Weekly|Primitive)Map` to get data for plotting.

### Embeds
Since data is loaded in `ExploreFocusSplit`, it is fully loaded by the time it reaches a plot component (previously the plots would load data themselves). Since it is now passed as a prop, all the raw data would be put into the URL by `Widget` and `ZodQueryEncoder`. To avoid that, some props can now be converted while being saved into the URL or loaded from it. This is done by making `Widget` async and replacing `ZodQueryEncoder` by `AsyncZodQueryEncoder`. 

Specifically, sample sets are passed to components like `variantSampleSet: SampleSetWithSelector`. In normal operation this is the prop the component is passed and also the prop it receives. No kind of conversion or loading is happening (same as with normal React components). When an embed link is generated, `variantSampleSet: SampleSetWithSelector` will be converted into `variantSampleSelector: NewSampleSelector` and saved into the URL (along with other props which don't need conversion, e.g. `country: string`). When an embed link is visited, `variantSampleSelector: NewSampleSelector` is used to load `variantSampleSet: SampleSetWithSelector`. Once all loading is complete, the component is rendered. This loading is the reason why `Widget` is now async.

Take a look at `VariantAgeDistributionPlotWidget` (only converted props) and `VariantInternationalComparisonPlotWidget` (props like `logScale` are not converted) for examples.

### Date cache
During implementation I noticed that date parsing and manipulation was taking ~80% of the all JS time. This is because we are getting ~20'000 entries from the new samples endpoint when showing a variant from a cold start (mainly `wholeSampleSet`). Since each unique date gets parsed around 100 times, I added a cache.

The cache is called `DateCache`. It is accessible as `globalDateCache`, but is normally called for you by `SampleSet`. Dates are stored as `UniqueDay` and `UniqueIsoWeek`. These contain all the most common forms of dates that we use. It is guaranteed that two of these objects are reference equal exactly if they represent the same date, so they can be used as keys for `Map` and `Set`.

### International comparison data is slow to load
`wholeInternationalSampleSet` takes around 20s to load from the API. Here's a [sample request](https://cov-spectrum.ethz.ch/api/resource/sample2?mutations=ORF8%3AQ27*%2CS%3AA570D%2CORF1a%3AT1001I%2CORF8%3AR52I%2CORF1a%3AG3676-%2CS%3AN501Y%2CS%3AT716I%2CORF8%3AY73C%2CORF1a%3AA1708D%2CN%3AS235F%2CORF1a%3AI2230T%2CORF1a%3AF3677-%2CS%3AS982A%2CS%3AY144-%2CN%3AD3L%2CORF1a%3AS3675-%2CS%3AH69-%2CS%3AD1118H%2CS%3AP681H%2CS%3AV70-&matchPercentage=0.8). I've made the international comparison plot load separately, so this isn't too much of a problem. The whole `DeepFocusPage` will wait until international comparison data is loaded (even if you want to view a different deep focus page) - this can be "fixed", but it will make the code more complicated.

### Zod is slow
I temporarily disabled Zod for the new sample set API calls, since it was really taking ages to validate the responses (much longer than date parsing). It turns out that Zod is actually [very slow](https://github.com/moltar/typescript-runtime-type-benchmarks), which is not something I thought about when I chose it. We have to switch to a different validation library. I opened issue #79 for that.

## Notes
This PR also closes #65, since I was making related changes anyway.


chaoran-chen:
Thanks, @tehwalris! I will test it in detail in the next few days.

Just a first quick note: The endpoint has the parameter `dateFrom` that could be used for the small preview plots of the known variants if we only need the data from the last two months.

tehwalris:
Thanks for that suggestion Chaoran. It turns out that that change doesn't improve the loading time much though:

saercig
srchoua

One of these is only requesting 3 months, the other is not limiting by time. It's hard to see which one is which.

chaoran-chen (approved):
Everything seems to work very well 👍 Thanks, @tehwalris!

tehwalris:
I'll merge this already, since merging this after other changes might be a lot of work. If you have any feedback @TKGZ, leave it on this PR and I'll make adjustments or open issues.

spacephoton:
Sounds good!

Issue #68, opened 2021-03-20
Improve the look of the country selection


chaoran-chen:
The country (and region) selection in the top bar deserves an improvement :)

See also #67

spacephoton:
Do you think a selection similar to comparison could work and be useful? 
![image](https://user-images.githubusercontent.com/36269621/117179512-7bb38780-add3-11eb-86f6-66205b50ac21.png)

I personally don't think it's useful for everything, so probably not worth investing the effort for it.

What improvements do you have in mind @chaoran-chen?

chaoran-chen:
I think that it is not the most important task right now, so I will move it back to the backlog.

chaoran-chen:
One reason to have an advanced country/region selection would be to make it clearer that it is possible to select a continent (and eventually the whole world). Maybe we can show something like this if the user clicks on the country selection? (The circles represent flags or a symbol for a region)

![image](https://user-images.githubusercontent.com/18666552/118054383-754e7e00-b386-11eb-9030-77287de6a99d.png)

And when the user starts to type, it will filter.

We could also additionally group the countries by continent. This way, a user with not the best geographic knowledge can easily click through the countries on a continent.

chaoran-chen:
The view is good enough for now.

PR #97, opened 2021-04-29
Regions - Continent view

### LINKED ISSUE #67: Global/continent view
Instead of a country, the user should be able to select a region (or continent) or the whole world. The names of the regions can be retrieved from `/resource/region`.

I think that the URL structure should stay as it is, i.e., `/explore/{country|region|"world"}/...`.

It depends on https://github.com/cevo-public/cov-spectrum-server/issues/11 and #62

spacephoton:
#67 

By default, a country can either be a region or a country. 
When api is called with country, it checks this with simple `isRegion` function to decide to pass it as a country or a region.

This was to keep things simple. 
If this becomes a problem long-term, then we can use a new type called "Place", but updating all the variables takes time, might be good to update when/if we remove zod. 

Also made some style fixes.

### DIFF
--- a/src/components/RequiredCountrySelect.tsx
+++ b/src/components/RequiredCountrySelect.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { Country } from '../services/api-types';
-import { CountrySelect } from './CountrySelect';
+import { PlaceSelect } from './PlaceSelect';
 
 export interface Props {
   id?: string;
@@ -13,13 +13,13 @@ export const RequiredCountrySelect = ({ id, selected, onSelect }: Props) => {
   const [visuallySelected, setVisuallySelected] = useState<Country | undefined>(selected);
 
   return (
-    <CountrySelect
+    <PlaceSelect
       id={id}
       selected={menuVisible ? visuallySelected : selected}
-      onSelect={country => {
-        setVisuallySelected(country);
-        if (country) {
-          onSelect(country);
+      onSelect={place => {
+        setVisuallySelected(place);
+        if (place) {
+          onSelect(place);
         }
       }}
       onMenuToggle={show => {
--- a/src/services/api.ts
+++ b/src/services/api.ts
@@ -371,9 +371,8 @@ export const getRegions = (): Promise<Region[]> => {
 export const getPlaces = async (): Promise<Place[]> => {
   const url = HOST + '/resource/country';
   const countries = await getCountries();
-  const regions = await getRegions();  
+  const regions = await getRegions();
   return countries.concat(regions);
-
 };
 
 export const fetchTimeDistributionData = () => {};
--- a/src/services/api.ts
+++ b/src/services/api.ts
@@ -5,6 +5,8 @@ import { defaultForNever, unreachable } from '../helpers/unreachable';
 import { AccountService } from './AccountService';
 import {
   Country,
+  Region,
+  Place,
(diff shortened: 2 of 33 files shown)


src/components/RequiredCountrySelect.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { Country } from '../services/api-types';
-import { CountrySelect } from './CountrySelect';
+import { PlaceSelect } from './PlaceSelect';
```
tehwalris, line 3: Since you renamed `CountrySelect` to `PlaceSelect`, you should also rename `RequiredCountrySelect` to `RequiredPlaceSelect`.

src/services/api.ts
```diff
@@ -341,7 +363,12 @@ export const getInterestingVariants = (
   },
   signal?: AbortSignal
 ): Promise<InterestingVariantResult> => {
-  const endpoint = `/computed/find-interesting-variants?country=${country}`;
+  let endpoint = `/computed/find-interesting-variants?`;
+  if (isRegion(country)) {
+    endpoint += `country=Switzerland`;
```
tehwalris, line 368: This should not be reachable, so this should throw. Defaulting to Switzerland could cause bugs in the future.

tehwalris (commented):
This seems to work fine. I left a few comments about small changes. Feel free to merge once those are done.

Overall treating `country` as place is a hack, like you said. It will be confusing in the future - especially since you can never be sure if a field is really a country or it could also contain a region. The "clean" option would be to use `place: Place` everywhere (rename all fields) - this would break embed URLs, so we can only do it now. I think that the hack is (borderline) worth it, since it is the lowest effort way to make the change. I'd like to hear the opinion of @chaoran-chen about that though.

I noticed there is no "world" support. Either do it as part of this PR or open an issue, so that we can close #67. 

The region selection is not that intuitive, since you have to know which region names are valid to search for them. For now it's great to have the feature at all though. This UX is something that we should improve with #68.

Please keep the unrelated style improvements separate next time, if it's not too much work. That way it's easier to see the relevant changes when reviewing. BTW the new confidential border is much better than the grey background.

chaoran-chen:
It would indeed be great if we can avoid breaking the embed links too often in the future. I expect the website to go online tomorrow (yeah!!) and then next week, I would like to embed a few widgets on other websites.

tehwalris:
> It would indeed be great if we can avoid breaking the embed links too often in the future. I expect the website to go online tomorrow (yeah!!) and then next week, I would like to embed a few widgets on other websites.

So do you mean we should change to `place` now or keep it called `country` for the foreesable future?

chaoran-chen:
I think that we can keep it. In most of the time, we will be using a country anyway so that the name fits well.

spacephoton:
> > It would indeed be great if we can avoid breaking the embed links too often in the future. I expect the website to go online tomorrow (yeah!!) and then next week, I would like to embed a few widgets on other websites.
> 
> So do you mean we should change to `place` now or keep it called `country` for the foreesable future?

Use place when the value could be either a country or a region. But for now, many countries may actually be regions but this is checked in the api call. This shouldn't break anything since whenever we pass a country it will still behave normally.

spacephoton:
Thanks @tehwalris @chaoran-chen for review! made chenges in my other pull request.

spacephoton:
Will make sure to keep important styling changes seperate :)

PR #99, opened 2021-04-29
Feature/Save as PNG or CSV

### LINKED ISSUE #48: Export plot as PNG
Hey. ~Since recharts generates SVG elements, would it be easy to export a plot as an SVG file?~ PNG is also good for now. Could we have it for all the plots?

The reason is again that I believe that it would be very helpful if the user can easily download the plots and use them in their papers and presentations.

spacephoton:
New wrapper component called `DownloadWrapper` that adds buttons on the top right to download wrapped chart as PNG image and/or as csv. #48


src/components/KnownVariantsList/29 april 
```diff
@@ -0,0 +1,29 @@
+
```
chaoran-chen: Hi. I guess that you uploaded this file by mistake. The filename has an empty space at the end which is not allowed in Windows so that I can't checkout the branch.

spacephoton: Accidently included my task list, sorry XD 

Fixed

chaoran-chen:
I'm just wondering if it might be better to integrate the `DownloadWrapper` into `NamedSection` and `WidgetLayout` and reuse the toolbar and make the buttons appear next to the "Show more" and "Share" buttons. We can also transform the "Share" button to an icon. Only "Show more" needs to stay as a big and text-filled button.

src/charts/DownloadWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const componentRef = useRef(null);
+
+  const exportOptions = {
+    fileName: name + '.png',
+    html2CanvasOptions: {
+      scale: 8,
+    },
+  };
+
+  //keep for debug
```
tehwalris: Please don't keep commented out code around for debugging. It often never gets used again. You can stash this code if you want to keep it for yourself.

src/charts/DownloadWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+  height: 100%;
+`;
+
+//Adds button to download wrapper component as an image
+const DownloadWrapper = ({
+  name = 'plot',
+  rawData = [],
+  dataProcessor = (data: any[]) => data,
+  ...props
+}: any) => {
```
tehwalris: Do not use `any` if you can reasonably avoid it. Is there some challenge writing the type in this case?

src/charts/DownloadWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+              onClick={downloadCSV}
+            />
+          </CSVLink>
+        )}
+        <DownloadButton
+          data-for='downloadPNG'
+          data-tip='Download this chart as PNG image.'
+          size={BUTTON_SIZE}
+          color='#95a5a6'
+          onClick={() => exportComponentAsPNG(componentRef, exportOptions)}
```
tehwalris, line 85: Just a heads up: Exporting "Hospitalization rates" on my machine takes around 8 seconds (production build). I don't think there's much we can do, and it's still worth it to keep the feature.

export-png-cpu

src/widgets/HospitalizationDeathPlot.tsx
```diff
@@ -153,18 +177,20 @@ export const HospitalizationDeathPlot = ({
...
-          total={total}
-          texts={makeTexts(variantName)[field]}
-          width={width}
-          height={height}
-          extendedMetrics={extendedMetrics}
-          onClickHandler={noopOnClickHandler}
-        />
-      )}
-    </div>
+    <DownloadWrapper name='HospitalizationDeathPlot' rawData={processedData} dataProcessor={dataProcessor}>
```
tehwalris, line 180: It would be simpler and less abstract if `DownloadWrapper` took `data` (which is directly the output of `dataProcessor(processedData)`), instead of taking `rawData` and `dataProcessor`. You can use `useMemo` to stop `dataProcessor` running on every render.

tehwalris (changes requested):
Overall the both PNG and CSV exports seem to work well. The code is clean and surprisingly simple! There are some minor things that I think should be changed (see comments).

tehwalris:
> I'm just wondering if it might be better to integrate the `DownloadWrapper` into `NamedSection` and `WidgetLayout` and reuse the toolbar and make the buttons appear next to the "Show more" and "Share" buttons. We can also transform the "Share" button to an icon. Only "Show more" needs to stay as a big and text-filled button.

I'd like to talk about this in today's meeting.

spacephoton:
Seems like the main issue here is when we want to pass the data to be
exported as csv

src/widgets/HospitalizationDeathPlot.tsx
spacephoton: How can we avoid making it update on every data change?

My idea was to make it only run when the person for sure wants to download the data, since it should be quite rare.

PR #109, opened 2021-04-30
Embed link goes to original page instead of home

### LINKED ISSUE #88: Embed-widget: better back-reference
Right now, when a widget is embedded on an external page, it links back to our homepage. It would be better if it references the page of a variant (in the case of a widget that shows information about a variant) or at least the explore page of a country (in the case of the sequencing intensity plot).

tehwalris:
This closes #88. The original page is now saved in the embed URL and the `CoV-Spectrum` link goes there. If the original page is unknown (typically because of an old URL), the link goes to the homepage like before.


src/widgets/Widget.tsx
```diff
@@ -4,30 +4,79 @@ import {
...
+import { AsyncQueryEncoder, AsyncZodQueryEncoder, MergedAsyncQueryEncoder } from '../helpers/query-encoder';
+import * as zod from 'zod';
+
+const SharedWidgetPropsSchema = zod.object({
+  originalPageUrl: zod.string().optional(),
+});
+
+type SharedWidgetProps = zod.infer<typeof SharedWidgetPropsSchema>;
+
+class SharedWidgetPropsEncoder implements AsyncQueryEncoder<SharedWidgetProps> {
```
tehwalris: This wrapper class handles compatibility with our old URLs, which don't have a `sharedWidgetJson` query parameter. We can remove this class if/once it's ok to break compatibility with URLs generated before this PR was merged. @chaoran-chen When should we remove this compatibility class?

chaoran-chen: You can remove it now :)

chaoran-chen (approved):
Works well. Thank you!

spacephoton:
Don't know enough about this to comment much, but seems good!

PR #110, opened 2021-04-30
Feature/new international plot

### LINKED ISSUE #90: Better country selection for the international comparison plot
Let's make the international comparison plot slightly better: Instead of the fixed few countries, it should show the currently selected country, Switzerland, and the three countries with the most number of absolute cases of the variant (excluding the currently selected country and Switzerland). Switzerland should always be red. The currently selected country - if it's not Switzerland - should also have a fixed color.

### LINKED ISSUE #96: Improve international comparison chart
Even though the international comparison plot is not fitting perfectly into the otherwise country-centric structure, it is extremely useful, and I believe that we are going to keep it this way for a while.

Therefore, I think that it's time to improve the layout and abandon Plotly for a Recharts-based implementation.

(See #90 for another issue regarding this plot.)

spacephoton:
#90 #96 

image
image
image
image

- By default, shows currently selected country along with top 4 with most samples
- Switzerland always red, current country blue, others random
- Press x to clear all countries except currently selected country (which cannot be cleared)


src/widgets/VariantInternationalComparisonPlot.tsx
```diff
@@ -1,19 +1,57 @@
...
+const DEFAULT_SHOW = 4;
+
+interface PlaceOption {
+  value: string;
+  label: string;
+  color: string;
+  isFixed: boolean;
+}
+
+const colourStyles: Partial<Styles<any, true, any>> = {
```
chaoran-chen: I think that we used US English so far ;)

src/widgets/VariantInternationalComparisonPlot.tsx
```diff
@@ -22,39 +60,94 @@ interface Props {
...
+
+const getPlacesMostVariantSamples = (
+  variantSamplesByPlace: Map<Place, any>,
+  exclude: Place,
+  n = DEFAULT_SHOW
+): string[] => {
+  const result = Array.from(variantSamplesByPlace)
+    .map((entry: [Place, unknown[]]) => ({
+      place: entry[0],
+      count: entry[1]?.length,
```
chaoran-chen, line 80: I believe that's wrong. I think that entry[1] is a `ParsedMultiSample` that represents (potentially) multiple samples. Instead of using the length of the sample set, it is necessary to sum up `entry[1].count`

src/widgets/VariantInternationalComparisonPlot.tsx
```diff
@@ -22,39 +60,94 @@ interface Props {
...
+  useEffect(() => {
+    const initialPlaces = [country].concat(getPlacesMostVariantSamples(variantSamplesByCountry, country));
+    const newOptions = initialPlaces.map((place: Place) => ({
+      value: place,
+      label: place,
+      color: getPlaceColor(place, country),
+      isFixed: place === country,
+    }));
+    setSelectedPlaceOptions(newOptions);
+  }, [country]);
```
chaoran-chen: My editor says:

`ESLint: React Hook useEffect has a missing dependency: 'variantSamplesByCountry'. Either include it or remove the dependency array.(react-hooks/exhaustive-deps)`

Does it make sense?

chaoran-chen (changes requested):

chaoran-chen:
The layout and multi-filter are fantastic!!

spacephoton:
Restyled the hospital slightly (sorry for not doing in seperate issue), and made the standard blue darker
image

PR #119, opened 2021-05-05
fix localstorage error

### LINKED ISSUE #117: Can't open embedded widgets in Chrome's incognito mode
In Chrome's incognito mode, it is not possible to load the CoV-Spectrum widgets that are embedded in another page (due to the security policies). See for example https://cevo-public.github.io/Quantification-of-the-spread-of-a-SARS-CoV-2-variant/

image

spacephoton:
fixes #117


src/services/AccountService.ts
```diff
@@ -86,4 +86,12 @@ export class AccountService {
   }
 }
 
+const getLocalStorageItem = (key: string): string | null => {
+  try {
+    return localStorage.getItem(key);
+  } catch (e) {
+    return null;
```
tehwalris, line 104: Print the error to the console instead of swallowing it.

src/services/AccountService.ts
```diff
@@ -86,4 +86,12 @@ export class AccountService {
   }
 }
 
+const getLocalStorageItem = (key: string): string | null => {
+  try {
+    return localStorage.getItem(key);
```
tehwalris, line 92: A comment like `May fail due to security policies in Chrome's incognito mode` would be helpful for future readers.

tehwalris (changes requested):
This seems like the right fix. Could `removeItem` also crash though? (triggered by a `logout` after `getJwt`) In that case please apply the same fix there too.

spacephoton:
> This seems like the right fix. Could `removeItem` also crash though? (triggered by a `logout` after `getJwt`) In that case please apply the same fix there too.

Fixed :)

tehwalris (approved):

Issue #34, opened 2021-03-02
Add Github mark


chaoran-chen:
A Github mark linking to this repository should be added - maybe to the footer?

spacephoton:
how about adding to the footer on the left since there is unused space?

chaoran-chen:
Makes sense!

chaoran-chen:
Now we don't have a footer anymore..

PR #121, opened 2021-05-05
Github buttons

spacephoton:
* about page
image

* left footer
image

Could do some prettifications but we can do that later. 

#34


spacephoton:
We could consider putting it on the website header but that might be overkill

chaoran-chen:
Hi @TKGZ! Thanks for your work and I'm very sorry for not being clearer but I really only thought about including the small Github mark/icon (see below) and link it to our repository but having buttons for starring + watching seems too much for me. The purpose of the mark is to let everyone know that it is open source and give them a place to submit issues but I don't consider stars as very important.

image

spacephoton:
No worries this only took a few minutes, where do you think it should be put?

chaoran-chen:
hmm. top-right? I'm really not sure..

spacephoton:
@chaoran-chen I think it is okay to put it in the about page, and nowhere else: 
image

spacephoton:
How about just keeping the small issue icon button, and removing the watch/star?

spacephoton:
image

spacephoton:
and maybe replace it with the github icon? 
image

chaoran-chen:
I think that's still too much. In my opinion, a simple mark is entirely sufficient.

spacephoton:
image
Small button only

package.json
```diff
@@ -17,6 +17,7 @@
     "@types/react-bootstrap": "^0.32.25",
     "@types/react-csv": "^1.1.1",
     "@types/react-dom": "^17.0.0",
+    "@types/react-github-button": "^0.1.0",
```
chaoran-chen: I guess that we won't need this then?

chaoran-chen (changes requested):
Thanks!

Please just remove the unused dependency. You can then merge :)

PR #124, opened 2021-05-07
Relative hospitalization

### LINKED ISSUE #100: Relative mode for hospitalization/death plots
As requested by Tanja "also show the change in probability for a variant (ie eg prob B117 hospitalization / prob nonB117 hospitalization)". I suggest doing the following: add a mode where the Y axis is normalized to "other variants". For example in the "80+" column we might have "other variants" at y=1 and B.1.1.7 at y=1.8. In this case we need to display the uncertainty of the ratio (see `approximateBinomialRatioConfidence` in branch `feature/relative-hospitalization`).

tehwalris:
This closes #100.

I ended up not hiding data points with very wide confidence intervals. Instead the max of the Y axis is set 2 times the largest Y data point (excluding error bars), so very wide error bars will be cut off. This is explained in the text below the plot and is also intuitively visible since the little horizontal bits of the bars are hidden in that case.

### DIFF
--- a/src/charts/GroupedProportionComparisonChart.tsx
+++ b/src/charts/GroupedProportionComparisonChart.tsx
@@ -109,7 +109,7 @@ export type Props = {
   width: number;
   height: number;
   extendedMetrics?: boolean;
-  rawY?: boolean;
+  maxY?: number; // percentage formatting is disabled if maxY is used
   hideReferenceScatter?: boolean;
   onClickHandler?: OnClickHandler;
 };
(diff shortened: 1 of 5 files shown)


spacephoton:
Not sure about putting so much text below, do you have other ideas of what we could do to explain what the data shows? @chaoran-chen @tehwalris Usually best to make graphs as self-explanatory as possible, such that only small explanation is necessary.

tehwalris:
> Not sure about putting so much text below, do you have other ideas of what we could do to explain what the data shows? @chaoran-chen @tehwalris Usually best to make graphs as self-explanatory as possible, such that only small explanation is necessary.

I agree that plots should be self-explanatory. I think the plots are already understandable without reading the text. The text is intended for someone who is presenting these plots to others, so that they can be confident about the details. For that reason I'll keep the text.

Since this looks like the only feedback, I'll merge now.

PR #125, opened 2021-05-09
Feature/estimated case numbers

### LINKED ISSUE #116: Estimated case numbers plot
As Tanja suggested, it would be useful to have a plot showing the estimated absolute number of cases of a variant. For example, for B.1.617 in the UK where the total case numbers are dropping but the proportion of B.1.617 is increasing, we would like to know how the total number of B.1.617 is developing.

I suggest the following way of calculation:

For a day _i_, let the number of all sequenced samples on day _i_ be _n_i_ and the number of samples of a variant among them be _k_i_. Further, let the number of all reported cases be _a_i_. Then, the estimated **smoothed** proportion of the variant on day _i_ is:

image

Finally, the estimated case number of a variant is _a_i * p_i_. We can compute a confidence interval for _p_i_ again by using the Wilson score, in the same way as for the hospitalization and death rate plots.

The total number of cases can be fetched from the endpoint for the sequencing intensity plot: https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md#sequencing-intensity-through-time

chaoran-chen:
As discussed, the sequencing intensity data is now loaded by the ExploreFocusSplit and passed down both to the explore and the focus page. The data is then used by the estimated case numbers plot. One open problem is #126.

Closes #116 

---

A screenshot :)

image

### DIFF
--- a/src/charts/EstimatedCasesChart.tsx
+++ b/src/charts/EstimatedCasesChart.tsx
@@ -1,0 +1,195 @@
+import React, { useMemo, useState } from 'react';
+import { UnifiedDay } from '../helpers/date-cache';
+import { ChartAndMetricsWrapper, ChartWrapper, colors, TitleWrapper, Wrapper } from './common';
+import { Area, ComposedChart, Line, ResponsiveContainer, Tooltip, XAxis, YAxis } from 'recharts';
+import Metric, { MetricsSpacing, MetricsWrapper } from './Metrics';
+import { getTicks } from '../models/wasteWater/WasteWaterTimeChart';
+import calculateWilsonInterval from 'wilson-interval';
+
+export type EstimatedCasesTimeEntry = {
+  date: UnifiedDay;
+  cases: number;
+  sequenced: number;
+  variantCount: number;
+};
+
+export type EstimatedCasesChartProps = {
+  data: EstimatedCasesTimeEntry[];
+};
+
+type PlotEntry = {
+  date: Date;
+  estimatedCases: number;
+  estimatedCasesCI: [number, number];
+};
+
+export function formatDate(date: number) {
+  const d = new Date(date);
+  return d.getDate() + '.' + (d.getMonth() + 1) + '.' + d.getFullYear();
+}
+
+const CHART_MARGIN_RIGHT = 15;
+
+export const EstimatedCasesChart = React.memo(
+  ({ data }: EstimatedCasesChartProps): JSX.Element => {
+    const [active, setActive] = useState<PlotEntry | undefined>(undefined);
+
+    const {
+      plotData,
+      ticks,
+    }: {
+      plotData: PlotEntry[];
+      ticks: number[];
+    } = useMemo(() => {
+      // Only show the data after the variant was first identified
+      const sortedData = [...data].sort((a, b) => (a.date.dayjs.isAfter(b.date.dayjs) ? 1 : -1));
+      const filteredData: EstimatedCasesTimeEntry[] = [];
+      let firstVariantFound = false;
+      for (let d of sortedData) {
+        if (!firstVariantFound) {
+          if (d.variantCount > 0) {
+            firstVariantFound = true;
+          }
+        }
+        if (firstVariantFound) {
+          filteredData.push(d);
+        }
+      }
+
+      const smoothedData: EstimatedCasesTimeEntry[] = [];
+      for (let i = 3; i < filteredData.length - 3; i++) {
+        const window = [
+          filteredData[i - 3],
+          filteredData[i - 2],
+          filteredData[i - 1],
+          filteredData[i],
+          filteredData[i + 1],
+          filteredData[i + 2],
+          filteredData[i + 3],
+        ];
+        const sum = (accumulator: number, currentValue: number) => accumulator + currentValue;
+        smoothedData.push({
+          date: filteredData[i].date,
+          cases: window.map(d => d.cases).reduce(sum),
+          sequenced: window.map(d => d.sequenced).reduce(sum),
+          variantCount: window.map(d => d.variantCount).reduce(sum),
+        });
+      }
+
+      const plotData: PlotEntry[] = [];
+      for (let { date, cases, sequenced, variantCount } of smoothedData) {
+        if (sequenced === 0) {
+          continue;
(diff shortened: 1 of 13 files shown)


chaoran-chen:
Hey. I will merge it because it seems to work quite well so far. But feel free to review anyway or create new issues if you see ways to improve!

src/charts/EstimatedCasesChart.tsx
```diff
@@ -0,0 +1,195 @@
...
+          }
+        }
+        if (firstVariantFound) {
+          filteredData.push(d);
+        }
+      }
+
+      const smoothedData: EstimatedCasesTimeEntry[] = [];
+      for (let i = 3; i < filteredData.length - 3; i++) {
+        const window = [
```
tehwalris, line 61: I think this might not work correctly if there are missing days. Ideally we would zero fill the data. If you think this is worth doing, I can give some implementation tips.

src/charts/EstimatedCasesChart.tsx
```diff
@@ -0,0 +1,195 @@
...
+                  domain={[
+                    (dataMin: any) => dataMin,
+                    () => data[data.length - 1].date.dayjs.toDate().getTime(),
+                  ]}
+                  ticks={ticks}
+                />
+                <YAxis />
+                <Tooltip
+                  active={false}
+                  content={e => {
```
tehwalris, line 138: It's kind of a hack to have side effects in a render function like this, but it seems to work fine, so I guess leave it.

src/charts/EstimatedCasesChart.tsx
```diff
@@ -0,0 +1,195 @@
...
+                  dot={false}
+                  isAnimationActive={false}
+                />
+              </ComposedChart>
+            </ResponsiveContainer>
+          </ChartWrapper>
+          <MetricsWrapper>
+            <MetricsSpacing />
+            <Metric
+              value={active !== undefined ? active.estimatedCases.toFixed(0) : 'NA'}
```
tehwalris, line 172: We use `-` instead of `NA` in the rest of the app, so please change it for consistency.

src/helpers/sequencing-intensity-entry-set.ts
```diff
@@ -0,0 +1,13 @@
...
+import { NewSampleSelector } from './sample-selector';
+
+export const SequencingIntensityEntrySetSelectorSchema = zod.object({
+  country: zod.string().optional(),
+});
+
+export type SequencingIntensityEntrySet = { data: SequencingIntensityEntry[] };
+export type SequencingIntensityEntrySetSelector = zod.infer<typeof SequencingIntensityEntrySetSelectorSchema>;
+export type SequencingIntensityEntrySetWithSelector = SequencingIntensityEntrySet & {
+  readonly selector: NewSampleSelector;
```
tehwalris, line 12: Why is this `NewSampleSelector` and not `SequencingIntensityEntrySetSelector`?

src/helpers/sequencing-intensity-entry-set.ts
```diff
@@ -0,0 +1,13 @@
+import { SequencingIntensityEntry } from '../services/api-types';
+import * as zod from 'zod';
+import { NewSampleSelector } from './sample-selector';
+
+export const SequencingIntensityEntrySetSelectorSchema = zod.object({
+  country: zod.string().optional(),
+});
+
+export type SequencingIntensityEntrySet = { data: SequencingIntensityEntry[] };
+export type SequencingIntensityEntrySetSelector = zod.infer<typeof SequencingIntensityEntrySetSelectorSchema>;
```
tehwalris, line 10: `SequencingIntensitySelector` might be a nicer name.

src/services/api.ts
```diff
@@ -334,23 +337,30 @@ export async function getInformationOfPangolinLineage(
...
-  signal: AbortSignal;
-}): Promise<SequencingIntensityEntry[]> => {
+export const getSequencingIntensity = (
+  selector: SequencingIntensityEntrySetSelector,
+  signal?: AbortSignal
+): Promise<SequencingIntensityEntrySetWithSelector> => {
   let url = HOST + `/plot/sequencing/time-intensity-distribution?`;
-  if (isRegion(country)) {
-    url += `region=${country}`;
+  if (!selector.country) {
```
tehwalris, line 345: Why not make `country` required in `SequencingIntensityEntrySetSelector`?

src/widgets/EstimatedCasesPlot.tsx
```diff
@@ -0,0 +1,68 @@
...
+      variantCount: 0,
+    });
+  }
+  const variantCounts = variantSampleSet.countByDate();
+  for (let [unifiedDay, count] of variantCounts.entries()) {
+    const entry = data.get(unifiedDay);
+    if (!entry) {
+      // TODO This should never happen and indicate an error with the data. Once we have a error reporting system,
+      //   this could be reported?
+      continue;
```
tehwalris, line 37: Yes, we could report this in the future. Non-fatal errors could be reported [like this](https://docs.sentry.io/clients/javascript/usage/#capturing-messages). Fatal errors (throwing and not catching nearby) will be reported automatically.

For now it would be nice to `console.warn` before `continue`.

src/widgets/SequencingIntensityPlot.tsx
```diff
@@ -38,47 +39,25 @@ const groupByMonth = (entries: SequencingIntensityEntry[]): TimeIntensityEntry[]
 
 const processData = (data: SequencingIntensityEntry[]): any => groupByMonth(data);
```
tehwalris, line 40: I know this wasn't your change, but we should not cast to `any` here. Remove the return type annotation so that `TimeIntensityEntry[]` is inferred.

tehwalris (commented):
The new plot is really interesting to have! The code is good. Most of my comments are about optional changes - it's up to you whether to address them.

src/charts/EstimatedCasesChart.tsx
chaoran-chen: The sequencing intensity endpoint will return values for every day from 1.1.2020 to the current day, so this isn't an issue here. But this is of course not a good general implementation for smoothing.

PR #128, opened 2021-05-13
Basic mobile support

### LINKED ISSUE #77: Menu breaks on small screens
Here's how our menu should look:

menu-ok

On small screens it does this:

menu-bad-1

On even smaller screens it does this:

menu-bad-2

### LINKED ISSUE #102: Don't use split screen on mobile
On small screens (especially small iPads or similar), our app is mostly usable except for the explore-focus split. We should split that into two separate pages on small screens. Adjusting the which `Route` components are shown and adding a "Back to explore" button on the focus page (when the device is small) should be enough.

tehwalris:
This closes #77 and #102. It fixes the hamburger menu and disables split-screen on mobile. It also removes the single "Explore" tab like we discussed, since that was related.

The filters (country, sampling strategy, time) are still in the main navbar/hamburger. It's technically slightly hard to move them onto a separate row like @TKGZ suggested, so we can make that change separately.

You'll need to use your phone in landscape so that the charts are reasonably big. We could move the "metrics" below the charts on small screens. I don't think that that's worth the effort and maintenance burden at the moment.


tehwalris:
Closes #102. (GitHub doesn't understand "and")

src/Header.tsx
```diff
@@ -7,6 +7,44 @@ import { AccountService } from './services/AccountService';
...
+`;
+
+const LeftNavGroup = styled.div`
+  @media (min-width: 768px) {
+    margin-left: 1.5rem;
+    margin-right: auto;
+  }
+`;
+
+const GisaidText = styled.span`
```
chaoran-chen: We can't change the way we show the GISAID note. We are obligated to write "Enabled by data from [GISAID logo]" by a written agreement.

chaoran-chen (changes requested):
Only a small change is needed, otherwise, it's great! Thanks!

src/Header.tsx
tehwalris: I changed it to never hide that text: 
gisaid-full-text

PR #145, opened 2021-05-28
Cleaner buttons

spacephoton:
Made the secondary buttons more discrete.

Before:
CleanShot 2021-05-28 at 18 15 12@2x

After:
CleanShot 2021-05-28 at 18 14 46@2x

### DIFF
--- a/src/charts/TimeChart.tsx
+++ b/src/charts/TimeChart.tsx
@@ -84,8 +84,6 @@ export const TimeChart = React.memo(
       <Wrapper>
         <TitleWrapper id='graph_title'>
           Proportion of all samples sequenced on week {currentData.yearWeek.split('-')[1]}
-          {', '}
-          {currentData.yearWeek.split('-')[0] + ' '}({currentData.firstDayInWeek})
         </TitleWrapper>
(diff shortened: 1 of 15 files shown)


src/charts/TimeChart.tsx
```diff
@@ -84,8 +84,6 @@ export const TimeChart = React.memo(
       <Wrapper>
         <TitleWrapper id='graph_title'>
           Proportion of all samples sequenced on week {currentData.yearWeek.split('-')[1]}
-          {', '}
-          {currentData.yearWeek.split('-')[0] + ' '}({currentData.firstDayInWeek})
```
chaoran-chen, line 88: Why did you decide to remove the year? (I generally prefer to have dates that are unambiguous.)

What's about a compromise: keep the year but remove the first-day-in-week?

spacephoton: It's because we now have both types shown below and it gets pushed to a second line which pushes the graph below for certain dates :)

spacephoton: ok, i'll add the year

PR #150, opened 2021-05-30
Color and minor fixes

spacephoton:
- Sequencing intensity more neutral color (red = bad)
CleanShot 2021-05-30 at 12 48 58@2x
We could make it also green since sequencing is probably good :)

- Made share button grey since most people won't use it (it's only embed)

- Fixed wastewater colors
CleanShot 2021-05-30 at 12 50 08@2x


chaoran-chen:
Gray on gray might be a little difficult to see. But I like the idea of green!

spacephoton:
Green did not look good, I put black and made the tick bold
@chaoran-chen 

CleanShot 2021-05-30 at 13 36 27@2x

chaoran-chen (approved):
Looks good!

PR #151, opened 2021-05-30
Combined export buttons

### LINKED ISSUE #105: Combine export buttons together
Combine the new icon buttons (CSV and PNG export) and the old "Share" button into one new "Export" dropdown button. This requires changing the structure around `Widget` and similar internal components.

### LINKED ISSUE #111: Make hospitalization/death rate plots sharable
Since the hospitalization/death rate plots are now online as well, they should become sharable.

tehwalris:
This PR combines the "Share" (embed), PNG download and CSV download buttons into one dropdown-button called "Export" (closes #105). The way this works is documented in [docs/export.md](https://github.com/cevo-public/cov-spectrum-website/blob/feature/combined-export-buttons/docs/export.md). You don't need to change anything about normal widgets for them to use the new button.

I tried to make the dropdown button look consistent with the style introduced by @TKGZ. That required some changes in `src/helpers/ui.tsx`. It's slightly mixed with `Dropdown` from Bootstrap for now, but that can be changed in the future.

I switched to different libraries for CSV export, since the export needs to be triggered via function call, but the library we had wants to render a component. BTW I noticed that we only have CSV export for the hospitalization and death plots.

This PR also makes the hospitalization and death plots embeddable (closes #111). This was the easiest way to add an export button to them. However it is possible to have an export button without `Widget.SharableComponent` (see [docs/export.md](https://github.com/cevo-public/cov-spectrum-website/blob/feature/combined-export-buttons/docs/export.md)).


spacephoton:
Seems to work well and the new CSV is great!

spacephoton (approved):

chaoran-chen (approved):
Thank you very much, @tehwalris!

It works perfectly and I like the architecture. Also, thanks for the documentation!

PR #157, opened 2021-06-03
Update index.css to show surveillance selector for smaller screens

### LINKED ISSUE #156: All samples vs Surveillance dropdown only shows for screen width > 1280px
I've noticed that the all samples vs surveillance dropdown disappears for screens with width < 1280 px.

This is very confusing to me as a user. I liked to select "surveillance" but when I checked out the website on a laptop screen I could no longer find the dropdown, at first thinking you had got rid of the feature.

Only later, back on an external monitor I noticed that the dropdown was back.

I don't think it's good to get rid of such an important feature depending on the width of the screen. Rather, the button should be shrunk in size. 

As a quick temporary fix, the breaking point should be reduced to 1180 rather than 1280px, because there's definitely enough space there. This should help show the selection on more laptop screens and thereby reduce the urgency of the problem.

Here some screenshots at various screen widths:
Status quo, it's showing for width > 1280px
image

Status quo for width < 1280px, no selection shown
image

Quick fix: there's enough space to show it even with screens > 1170px
image

corneliusroemer:
Fixes #156

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -55,3 +55,9 @@ code {
 .custom-narrow-dropdown button.dropdown-toggle::after {
   vertical-align: 0.15em;
 }
+
+/* HACK Temporary fix to show surveillance selection dropdown on smaller laptop screens */
+@media (min-width: 1170px)
+.hidden.xl\:block {
+   display: block!important;
+}
--- a/src/index.css
+++ b/src/index.css
@@ -57,7 +57,8 @@ code {
 }
 
 /* HACK Temporary fix to show surveillance selection dropdown on smaller laptop screens */
-@media (min-width: 1170px)
-.hidden.xl\:block {
-   display: block!important;
+@media (min-width: 1170px) {
+  .hidden.xl\:block {
+    display: block !important;
+  }
 }


src/index.css
```diff
@@ -55,3 +55,9 @@ code {
 .custom-narrow-dropdown button.dropdown-toggle::after {
   vertical-align: 0.15em;
 }
+
+/* HACK Temporary fix to show surveillance selection dropdown on smaller laptop screens */
+@media (min-width: 1170px)
+.hidden.xl\:block {
+   display: block!important;
+}
```
chaoran-chen:
```suggestion
@media (min-width: 1170px) {
  .hidden.xl\:block {
    display: block !important;
  }
}
```

Hi @corneliusroemer, thanks for pointing out the issue and suggesting the temporary fix! I think that your code misses two braces but otherwise I agree with the changes.

chaoran-chen (changes requested):

corneliusroemer:
Preview doesn't work yet, will investigate, maybe need to tweak the css a bit more.

chaoran-chen (approved):
Sorry, I had to authorize the preview. Now it works. Thaks again!

PR #161, opened 2021-06-06
Basic world

### LINKED ISSUE #101: Add world compatibility

spacephoton:
CleanShot 2021-06-06 at 14 01 45@2x

Uses switzerland's interesting variants for now

### DIFF
--- a/src/components/KnownVariantsList/KnownVariantCard.tsx
+++ b/src/components/KnownVariantsList/KnownVariantCard.tsx
@@ -104,7 +104,7 @@ export const KnownVariantCard = ({ name, chartData, recentProportion, onClick, s
         )}
       </Title>
       <div>
-        <SimpleAreaPlot data={chartData} selected={selected} />
+        <SimpleAreaPlot data={chartData} selected={selected ? selected : false} />
       </div>
     </Card>
   );
--- a/src/components/PlaceSelect.tsx
+++ b/src/components/PlaceSelect.tsx
@@ -8,7 +8,7 @@ export interface Props {
   id?: string;
   selected: Place | undefined;
   onSelect: (country: Place | undefined) => void;
-  onMenuToggle?: (show: boolean) => void;
+  onMenuToggle: (show: boolean) => void;
 }
 
 export const PlaceSelect = ({ id, selected, onSelect, onMenuToggle }: Props) => {
@@ -18,6 +18,7 @@ export const PlaceSelect = ({ id, selected, onSelect, onMenuToggle }: Props) => 
     let isSubscribed = true;
     getPlaces().then(places => {
       if (isSubscribed) {
+        places.push("World");
         setPlaces(places);
       }
     });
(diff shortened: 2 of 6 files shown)


chaoran-chen:
I'll look into it more carefully later/tomorrow but here's a first bug:

image

Also for this endpoint, no parameter should be set for "world".

corneliusroemer:
Isn't it a bit misleading to just add all sequences up? It's not really "world" but 40% UK, 40% US, 5% Denmark, 5% Germany, 8% other wealthy countries and 2% rest of world.

Have you thought about weighting by population or, possibly better, incidence?

spacephoton:
Yes that is actually very misleading

I am not very sure how to solve.

corneliusroemer:
Weighting by population is probably not good.

I would suggest weighting by cases. Our World in Data tells us how many infections per country. Use sequencing results to turn them into lineage cases per country, then add them all up.

Problem: Many countries have 0 sequences in many weeks, in particular recent weeks. What do we do with them? Leave them out altogether? Issue would be, if a country like India which dominates global cases has only 1 sequence in a week, that single sample totally dominates the world plot. But that's the way it is, as long as the uncertainty is made clear. Better than totally biasing towards the few countries that sequence the most.

chaoran-chen:
World means that we include all submitted sequences in the world. We don't say that the samples are representative in any way and it would be difficult to decide when a sample set is really representative. The problem is not new and also exists on the national level. As we can see here (https://cov-spectrum-d-git-feature-sequencing-and-metadata-cover-352fb2.vercel.app/explore/Switzerland/AllSamples/AllTimes/sequencing-coverage), a lot of the data for Switzerland come from only a few cantons (maybe we have almost random mixing in Switzerland, maybe not..?). In other countries, it is probably even worse (e.g., many countries preferentially sequenced potential B.1.1.7 cases last winter..).

I suggest that we introduce the "world" option without any weighting. We can add more explanations about how the plots should be interpreted. We could try to develop a model to estimate the prevalence of a variant in the whole world, but that would be an entirely different and much larger task.

spacephoton:
@chaoran-chen anything else to include? fixed pangolin part

chaoran-chen (changes requested):
Hi @TKGZ, I found two more calls where country=World is being used:

- [x] Fitness advantage:

image

- [x] And when you are on the sequencing coverage (deep explore) page (https://cov-spectrum-develop-lzsuyuzhm-cov-spectrum.vercel.app/explore/World/AllSamples/AllTimes/sequencing-coverage), the metadata availability plot makes its own call to the `/resource/sample2` call.

image

- [x] The if-is-region-or-world construct is duplicating quite often. Could we move the logic into a new function? E.g., one which takes an `URLSearchParams` and a `Place` and appends the correct values to the `URLSearchParams`?

spacephoton:
Could you have one more look @chaoran-chen :)

chaoran-chen:
Hey. The known variants list is not loading for me for world:

image

And I see that an invalid request is being made:

image

spacephoton:
@chaoran-chen 
CleanShot 2021-06-24 at 13 04 48
sample not returning anything when no region or country is set

for now, I will just set it to europe as a hack

chaoran-chen:
https://cov-spectrum.ethz.ch/api/resource/sample2?matchPercentage=1&dateFrom=2021-03-22&pangolinLineage=AY.1

I don't understand the problem. The above link works perfectly? And wouldn't setting it to Europe often generate very wrong results?

spacephoton:
That's strange, I'll have a look and fix

spacephoton:
Strange, I put it back the way it is before and now it works lol
:)

PR #163, opened 2021-06-07
Feature/sequencing and metadata coverage

### LINKED ISSUE #154: Sequencing representativeness in Switzerland
**Aim**

I'm thinking about adding a deep focus page to the sequencing intensity box for Switzerland. The sequencing intensity plot shows us how consistent we sequence through time. The aim of the new widget is to help us analyze whether we sample representatively with respect to the cantons, age, and sex.

It should allow some slightly deeper analysis and, for example, not only answer the question "How representative did we sample w.r.t. the cantons?" but also "How representative did the cantons sample w.r.t. different age groups?"

**Idea**

An idea is to have the attributes division, age, and sex as checkboxes and the user can choose only one...

image

or multiple:

image

This will of course create a lot of bars but I think that it is still easy to understand and to use. If we allow all three to be checked, it would create about 500 rows.

**API**

(The required endpoints are not there yet but we have the data and can create them once we decided how the plot should look like.)

(@TKGZ, what do you think?)

### LINKED ISSUE #162: Metadata completeness
Before viewing and "trusting" the plots that largely depend on the availability of metadata (e.g., the age and geographic distribution plots and hospitalization probabilities), it is important to know how complete the dataset is. Thus, we should have a plot showing the proportion of sequences for which we know the division, age, sex, hospitalization, and death status.

chaoran-chen:
This PR adds a "deep" page which is linked from the "show more" button of the sequencing intensity plot. Because the `DeepFocusPage` is variant-specific, I introduced the `DeepExplorePage`.

The aim of the new `SequencingCoverageDeepExplore`-page is to give an overview of the quality/completeness/representativeness of the data. It contains the `MetadataAvailabilityPlot` which is available for all countries (closes #162) and the `SequencingRepresentativenessPlot` which is only available for Switzerland (closes #154).

image

### DIFF
--- a/src/pages/DeepExplorePage.tsx
+++ b/src/pages/DeepExplorePage.tsx
@@ -6,18 +6,26 @@ import { Link } from 'react-router-dom';
 import React from 'react';
 import { Route, useRouteMatch } from 'react-router';
 import { SequencingCoverageDeepExplore } from '../components/SequencingCoverageDeepExplore';
+import { SequencingIntensityEntrySetWithSelector } from '../helpers/sequencing-intensity-entry-set';
 
 interface Props {
   country: Country;
   dateRange: DateRange;
   samplingStrategy: SamplingStrategy;
+  sequencingIntensityEntrySet: SequencingIntensityEntrySetWithSelector;
 }
 
 const routes: DeepRoute<Props>[] = [
   {
     key: 'sequencing-coverage',
     title: 'Sequencing Coverage',
-    content: props => <SequencingCoverageDeepExplore country={props.country} dateRange={props.dateRange} />,
+    content: props => (
+      <SequencingCoverageDeepExplore
+        country={props.country}
+        dateRange={props.dateRange}
+        sequencingIntensityEntrySet={props.sequencingIntensityEntrySet}
+      />
+    ),
   },
 ];
 
--- a/src/pages/DeepExplorePage.tsx
+++ b/src/pages/DeepExplorePage.tsx
@@ -4,7 +4,8 @@ import { DeepRoute, makeLayout, makeSwitch } from '../helpers/deep-page';
 import { Button } from 'react-bootstrap';
 import { Link } from 'react-router-dom';
 import React from 'react';
-import { useRouteMatch } from 'react-router';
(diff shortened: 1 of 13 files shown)


spacephoton:
Do you think we can make it without scrollbar and have it extend to the bottom instead?

CleanShot 2021-06-07 at 18 43 53@2x

chaoran-chen:
I think that it is nice that the left and right parts don't move. It becomes more important if the user selects both division and age group, then the plot gets really long.

src/pages/DeepExplorePage.tsx
```diff
@@ -0,0 +1,56 @@
...
+        samplingStrategy={props.samplingStrategy}
+        sequencingIntensityEntrySet={props.sequencingIntensityEntrySet}
+      />
+    ),
+  },
+];
+
+export const DeepExplorePage = (props: Props) => {
+  const { path, url } = useRouteMatch();
+  const _makeLayout = (content: JSX.Element) =>
```
spacephoton, line 35: why does this have an undersscore?

spacephoton (approved):
Ok, makes sense, since some lists are very long/

spacephoton:
Seems great!! :D

spacephoton:
Maybe a subtitle explaining a little bit more here?
CleanShot 2021-06-07 at 18 49 44@2x

spacephoton:
Because list is sometimes long, maybe we can put x axis above?

spacephoton:
Maybe useful to also sort by cases?

spacephoton:
Or have that option

src/pages/DeepExplorePage.tsx
chaoran-chen: Because it is importing a function called `makeLayout` from `'../helpers/deep-page'` (line 3)

chaoran-chen:
> Maybe a subtitle explaining a little bit more here?

Done

> Because list is sometimes long, maybe we can put x axis above?
> Maybe useful to also sort by cases? Or have that option

Good ideas. Let's address them in future PRs.

spacephoton:
Awesome!

Issue #178, opened 2021-06-24
Make short bars in bar plots tappable above bar for mobile usability


corneliusroemer:
Short bars are currently not easily tappable, because one has to hit the small rectangle both horizontally and vertically.

It would be good, if it was possible to select bars at full height, no matter their height.

See screencast for illustration. Otherwise, mobile UX is already great!

https://user-images.githubusercontent.com/25161793/123266334-302c7880-d4fc-11eb-8058-51ca50ae518b.mp4

spacephoton:
Thanks @corneliusroemer! The way they were implemented means there isn't an obvious way, but I will have another look :)

chaoran-chen:
https://github.com/cevo-public/cov-spectrum-website/blob/92efa86ce8fe7128e9ccbc1adc0ce9be25b422e5/src/widgets/MetadataAvailabilityPlot.tsx#L89-L99

I found a way using the tooltip component (see code above). [See here (Metadata availability).](https://cov-spectrum.ethz.ch/explore/Switzerland/AllSamples/AllTimes/sequencing-coverage)

spacephoton:
Woah @chaoran-chen !! I remember spending a lot of time a while ago and not figuring this out. Thanks!

spacephoton:
@corneliusroemer fixed now with #180 ! :)

corneliusroemer:
Wow, this is great! Thanks :)

PR #180, opened 2021-06-24
basic touchable intensity

spacephoton:
#178

@chaoran-chen before I continue, do you see any issues for sequencing intensity? 

I will probably have to change the way the tick is implemented since no longer using the index. Wasn't working so well anyway.

### DIFF
--- a/src/charts/Metrics.tsx
+++ b/src/charts/Metrics.tsx
@@ -95,17 +95,23 @@ export type MetricProps = {
 interface ChartAndMetricsProps {
   children: React.ReactNode;
   metrics: MetricProps[];
+  title: string;
 }
 
-export const ChartAndMetrics = ({ children, metrics }: ChartAndMetricsProps) => {
+export const ChartAndMetrics = ({ children, metrics, title }: ChartAndMetricsProps) => {
   return (
-    <div id="chart-and-metricss" className="flex flex-col lg:flex-row h-full">
-      <div className="w-full h-full">{children}</div>
-      <MetricsWrapper>
-        {metrics.map(mProps => (
-          <Metric {...mProps} />
-        ))}
-      </MetricsWrapper>
+    <div id='chart-and-metrics' className="flex flex-col h-full w-full">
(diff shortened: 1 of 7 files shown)


spacephoton:
Maybe also a good chance to simplify the implementation of the charts, I feel like they are unecessarily complex, and try to make them more reusable

chaoran-chen:
So far, it seems to work great, as long as we can bring the "changing ticks" back.

spacephoton:
@chaoran-chen can you have a look now? :)

spacephoton:
I also generalized part of the chart so they're easier to implement in future

src/charts/Metrics.tsx
```diff
@@ -92,6 +92,30 @@ export type MetricProps = {
...
+interface ChartAndMetricsProps {
+  children: React.ReactNode;
+  metrics: MetricProps[];
+  title: string;
+}
+
+export const ChartAndMetrics = ({ children, metrics, title }: ChartAndMetricsProps) => {
+  return (
+    <div id='chart-and-metrics' className='flex flex-col h-full w-full'>
+      <h3 id='chart-title' className='my-0 pb-4 pr-10 pt-0 text-gray-500'>
```
chaoran-chen: Why do we need to set an id?

Also a more general question: Shouldn't an id be unique in an HTML document and whenever we have multiple ChartAndMetrics loaded, wouldn't it be violated? Or does React have some kind of encapsulations in place?

chaoran-chen (changes requested):
Hi @TKGZ! Works nicely and it's great to add a feature while reducing the amount of code!

Just one thing that I think should be improved:

image

This looks a little crowded and the hole at the end looks weird. Further, for consistency, it would be better to write the year with four digits.

src/charts/Metrics.tsx
spacephoton: I actually don't know much about ID's, I just put them because they might be helpful later if we need to run pupeteer tests

spacephoton:
CleanShot 2021-07-01 at 13 13 28
Couldn't find elegant way to adjust tick distance with custom ticks, so made the date appear only when it is active.

PR #196, opened 2021-07-07
Feature/advanced search

### LINKED ISSUE #35: Combined pangolin lineage and mutation search
## Update (03.07.2021):

Let's merge the pangolin lineage search with the mutations search!

- It should be possible to put in at most one and an arbitrary number of amino acid mutations. 
- How to deal with "Match Percentage"?? Maybe let's leave it out in the merged search and keep the "Search by mutations" in the private version until we found a better solution. Let's see if someone misses it.
- The search bar should work similarly to the one in the international comparison plot (see screenshot below): the entries should be parsed in real-time. The user should get feedback when they enter an invalid value.

image

---

## Original text:

Hi. I received a request from @tanja819 earlier:

> Did we so far see any B117 which carry also a 484 mutation in Switzerland? It would be great to check for that in future regularly, as the 484 may induce immune escape

Do you think that we should be able to answer this type of request with Spectrum? If yes, how?

Right now, we tend to call a known variant if 80% of the lineage-defining mutations are present. One main reason for that is that sequencing is not perfect and we don't get full coverage so that we are not always sure about whether a mutation is present. Furthermore, missing a few mutations might not change the properties of a variant entirely. However, as in the case Tanja mentioned, there are certain mutations of special interest.

Should we implement an advanced search where the user can enter required and optional mutations?

### LINKED ISSUE #173: [Feature request] Allow immediate searching by pressing enter without requiring arrow down key
When I enter a lineage like `B.1.1.7` in the search bar, I can't just press `Enter` to submit. This is contrary to usual conventions and reduces usability.

Two possible fixes (if you want to):
a. Do an implicit down arrow when enter is pressed in the search bar
b. Explicitly select the first suggestion by default and underline the first suggestion in blue

image

### LINKED ISSUE #174: [Feature request] Explicitly suggest wildcard queries in the suggested search queries
Even though wildcards are valid search queries, they are never included in the suggested search queries. This harms discoverability and potentially reduces usability.

To not overload the suggested queries with `*` duplicates, one could just add the `*` as an additional option after the first suggested query like such:

Query typed: B.1.617
- B.1.617
- B.1.617*
- B.1.617.1
- B.1.617.2
- ...

This is status quo:
image

Wildcard queries are treated as non-existent:
image

chaoran-chen:
This closes #35, closes #173, closes #174.

Open tasks: #197, #198

Screenshot:

image

@TKGZ, @corneliusroemer - please give feedback! :)

### DIFF
--- a/src/components/KnownVariantsList/KnownVariantsList.tsx
+++ b/src/components/KnownVariantsList/KnownVariantsList.tsx
@@ -12,7 +12,6 @@ import {
   loadKnownVariantSampleSets,
 } from './load-data';
 import dayjs from 'dayjs';
-import { Typeahead } from 'react-bootstrap-typeahead';
 import _VARIANT_LISTS from './variantLists.json';
 import { KnownVariantsListSelection } from './KnownVariantsListSelection';
 import { formatVariantDisplayName } from '../../helpers/variant-selector';
@@ -44,10 +43,6 @@ export type VariantList = {
   source?: string;
   fillUpUntil: number;
 };
-
-const SearchWrapper = styled.div`
-  margin-bottom: 10px;
-`;
 
 function selectPreviewVariants(
   definedVariants: VariantSelector[],
@@ -85,12 +80,6 @@ export const KnownVariantsList = ({
 }: Props) => {
   const [selectedVariantList, setSelectedVariantList] = useState(VARIANT_LISTS[0].name);
   const [variantSampleSets, setVariantSampleSets] = useState<KnownVariantWithSampleSet<VariantSelector>[]>();
-  const [pangolinLineages, setPangolinLineages] = useState<
-    {
-      pangolinLineage: string;
-      count: number;
-    }[]
-  >([]);
   const [knownVariantSelectors, setKnownVariantSelectors] = useState<VariantSelector[]>([]);
 
   const knownVariantsWithoutData: {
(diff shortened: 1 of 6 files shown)


corneliusroemer:
Cool! Works great! Just a few cosmetic points you may already be aware of but just in case:
- Pango lineage when selected has red tint, looks like error, maybe make blueish
- Can select lineage by pressing enter, but another enter without any new character should submit query
- HTML markup lacks "input" tag for a18n and extensions such as vimium
- Would be nice to capitalise all characters for consistent look, even if entered lower case
- Would be cool, if WHO variants could be searched for and aliased, similar to how the asterisk * is

chaoran-chen:
@corneliusroemer, thanks for the input!

> - Pango lineage when selected has red tint, looks like error, maybe make blueish

Sure, I changed it to a blue for now. (I assume that @TKGZ is going to change the colors anyway ;))

> - Can select lineage by pressing enter, but another enter without any new character should submit query

See #202

> - HTML markup lacks "input" tag for a18n and extensions such as vimium

I see an input tag?

> - Would be nice to capitalise all characters for consistent look, even if entered lower case

See #204 

> - Would be cool, if WHO variants could be searched for and aliased, similar to how the asterisk * is

See #203

corneliusroemer:
>> HTML markup lacks "input" tag for a18n and extensions such as vimium

> I see an input tag?

Hmm, this is how it looks for me on Brave

image

image

This is how it should look like:

image

chaoran-chen:
The input is further down in the tree:

image

corneliusroemer:
Now I found it too, but it's very hidden. And Vimium doesn't pick it up, I'm wondering whether that's because it's buried.

PR #210, opened 2021-07-19
Division maps

### LINKED ISSUE #179: Division maps
We talked about it so often but apparently, we never created an issue for it?

It would be great to have a map to visualize the prevalence (or absolute numbers) of the divisions in a selected country. It would be a replacement/alternative to the current geography table.

As for the map type, I like what the BAG uses on their dashboard: https://www.covid19.admin.ch/de/repro/val

image

Finding the map data for all countries is certainly difficult but starting with a few countries and making the implementation easily extensible would be enough for now. For example, Switzerland, United States, and Germany (these are the three countries from which most of our users are from),

spacephoton:
Top three countries should work.

Please give me some feedback before I add more countries.

- How is the performance? 
- How can I make the UI more intuitive? 
- Why is the division data so delayed compared the current data?
- Do you know how I might be able to lazy load the maps?


spacephoton:
CleanShot 2021-07-19 at 21 32 53@2x

spacephoton:
CleanShot 2021-07-19 at 21 35 16@2x

chaoran-chen (changes requested):
This looks super cool!! I like it very much! The performance is great on my laptop (which however is probably above average) and I find the UI intuitive. Unfortunately, I don't have an answer to your two questions right now but we can talk about it on Friday if you like.

There are just a few issues/bugs/wishes:

- [x] I believe that you forgot to multiply the proportions by 100?
- [x] When hovering over divisions with, I guess, a proportion of zero (entirely white), it does not become "active".
- [x] The number of divisions with data should only count those divisions that are also visible on the map. Those that we cannot map should be excluded. Having 30 divisions with data in Switzerland for B.1.1.7 is a bit weird.
- [x] Would it make sense to limit the maximal height? The map of Germany seems to be a bit big to me.
- [x] In our API dataset (which we cannot change easily), we write `Niedersachsen` but the map data use `Lower Saxony`.
- [x] Could you please add some documentation in `docs/` explaining (1) how the map generally works and (2) the steps to add data for more countries?

spacephoton:
@chaoran-chen could you have a look at the fixes please? :)

spacephoton:
Still seems like the division data is delayed, for example the range for delta is < 12 % in this map
image

chaoran-chen:
Why do you think that <12% is wrong in the view that you selected? (Reminder: The proportion is not only using data from the last week/last few weeks but whatever is selected in the time range filter in the header.)

chaoran-chen (changes requested):
image

For the US, the number of divisions seems to always be zero.

Am I understanding it correctly that the "number divisions" metric currently just shows the number of divisions that a country has? Is that useful? What's about showing the number of divisions in which the selected variant was detected at least once in the selected time range? I imagine that the number is useful for a new variant when that number can help us assess how widely it has spread.

Otherwise, everything looks great!

spacephoton:
Oh! The time range! I see. I think this might be confusing then, what do you think of including the time range in the subtitle of the chart?

chaoran-chen:
Sure, we can do that.

spacephoton:
> image
> 
> For the US, the number of divisions seems to always be zero.
> 
> Am I understanding it correctly that the "number divisions" metric currently just shows the number of divisions that a country has? Is that useful? What's about showing the number of divisions in which the selected variant was detected at least once in the selected time range? I imagine that the number is useful for a new variant when that number can help us assess how widely it has spread.
> 
> Otherwise, everything looks great!

There is a problem because sometimes

spacephoton:
Finally beat procrastination and merged this one :)

PR #214, opened 2021-07-25
fix heatmap scroll

### LINKED ISSUE #195: Wastewater heatmap: Fix left column when scrolling
The left column with the mutation names of the wastewater heatmap should be fixed during horizontal scrolling if the width of the screen is large enough. 

image

spacephoton:
#195

When screen id bigger than tailwind `md`, keep left bar static, while the heatmap still scrollable horizontally.
CleanShot 2021-07-25 at 17 02 46@2x


chaoran-chen:
It would be great if you could fix the code style so that we can use the Vercel preview :)

chaoran-chen:
Hi @TKGZ. The labels are not aligned with the colored cells (in Chrome and Firefox on Windows). Do you (or @ningxie1991) have an idea how to fix it? Since the scrollbar has different heights in different browsers, I guess that this issue is not entirely trivial..

image

spacephoton:
> Hi @TKGZ. The labels are not aligned with the colored cells (in Chrome and Firefox on Windows). Do you (or @ningxie1991) have an idea how to fix it? Since the scrollbar has different heights in different browsers, I guess that this issue is not entirely trivial..

Ah yes, this is a problem to fix

spacephoton:
Yes, it depends on the browser. I added scrollbar on the left part too, can you have a look how it is now? On mac it's not a problem since it doesn't take space.

image

chaoran-chen (approved):
Lucky Mac :P

Now it works on Windows as well 👍 
image

spacephoton:
> Lucky Mac :P
> 
> Now it works on Windows as well 👍
> image

not the most pretty but it works!

PR #215, opened 2021-07-25
Search on enter + move button right

### LINKED ISSUE #202: Initiate search with enter
The user should be able to perform a search by clicking on Enter.

I already tried to implement it yesterday. There is an `onKeyDown` event handler that can be used to get the Enter event. However, it wasn't obvious to me how we can find out whether the user clicked on Enter to select a suggestion or not.

(Thanks @corneliusroemer for the suggestion.)

spacephoton:
CleanShot 2021-07-25 at 17 12 03@2x
#202


chaoran-chen (approved):
Looks good to me!

PR #218, opened 2021-07-30
Added search by comma separated list feature

### LINKED ISSUE #198: Search: Accept comma-separated list
The user should be able to copy&paste a comma-sperated list of mutations into the search field. The list should then be parsed.

In pictures:

image

to

image

Open question: What if one of the mutations is in an invalid format?

### LINKED ISSUE #216: Variant search: "remove all" button not working property
The "x" at the right side of the variant search field to remove all input is not working properly. 

image

How to reproduce the error:

1. Open https://cov-spectrum.ethz.ch/
2. Enter in the variant search field "B.1.1.7"
3. Click on the remove all "x" button
4. Enter S:N501Y
5. Click on search

Instead of showing the page for the "S:N501Y" variant, it shows "B.1.1.7 + S:N501Y".

ningxie1991:
For issue #198

1. Modified the AsyncSelect component by manually setting value, inputValue and menuIsOpen
2. Added handleInputChange, handleCommaSeparatedInput methods to handle the case when inputs have commas
3. User can directly paste a comma separated list of mutations (spaces will be trimmed). The list gets transformed into valid tags immediately, compliant to the rules and logic defined in the [suggestOptions ](https://github.com/cevo-public/cov-spectrum-website/blob/601be84331101ef9d1f497fa1b41ddca7e332be5/src/components/VariantSearch/VariantSearch.tsx#L108-L125) method.
4. User can also create valid tag by entering a mutation name then comma one by one (it behaves like the enter key) 

copy paste comma separated list

The following gif shows when there are invalid entries from the list. The valid mutation queries will be transformed into tags and the invalid queries text will remain in the input field and the drop down option "No Options" will be shown.  (I think this could be better and cleaner than showing an error message for those invalid queries.)

copy paste comma separated list with invalid queries

The search on enter feature still works.

@TKGZ @chaoran-chen Please review the code and test the features. Thanks!

### DIFF
--- a/src/components/VariantSearch/VariantSearch.tsx
+++ b/src/components/VariantSearch/VariantSearch.tsx
@@ -214,7 +214,7 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
 
   /**
    * This is needed because the code is converted to manually opening/closing the option menu,
-   * which overrides the default behaviour of the DropdownIndicator. 
+   * which overrides the default behaviour of the DropdownIndicator.
    * @param props props of component
    * @constructor the constructor
    */
--- a/src/components/VariantSearch/VariantSearch.tsx
+++ b/src/components/VariantSearch/VariantSearch.tsx
@@ -112,9 +112,6 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
   };
 
   const suggestOptions = (query: string): SearchOption[] => {
-    if(query.slice(-1) == ","){
-      query = query.slice(0, -1)
-    }
     const onePLAlreadySelected =
         selectedOptions.filter(option => option.type === 'pangolin-lineage').length > 0;
     const suggestions: SearchOption[] = [];
@@ -184,7 +181,8 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
   const promiseOptions = (inputValue : string) => {
     // resets the options to default (where input value is '') when menu is closed
     if(!menuIsOpen){
-      setInputValue('');
+      inputValue = ''
+      setInputValue(inputValue);
     }
 
     return Promise.resolve(suggestOptions(inputValue));
--- a/src/components/VariantSearch/VariantSearch.tsx
+++ b/src/components/VariantSearch/VariantSearch.tsx
@@ -175,7 +175,7 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
     setInputValue(invalidQueries);
     // leave options menu open if there are invalid queries from the list
     // this will show the "No Options" in the menu
-    setMenuIsOpen(invalidQueries === '' ? false : true);
+    setMenuIsOpen(invalidQueries !== '');
   }
 
   const promiseOptions = (inputValue : string) => {
--- a/src/components/VariantSearch/VariantSearch.tsx
+++ b/src/components/VariantSearch/VariantSearch.tsx
@@ -154,21 +154,24 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
       // remove spaces from both ends of the query
       query = query.trim()
       const suggestions = suggestOptions(query);
-      const selectedOption = suggestions.find(item => item.value.toUpperCase() === query.toUpperCase());
+      const selectedOption = suggestions.find(option => option.value.toUpperCase() === query.toUpperCase());
       // If query gets back valid a valid suggest option which is not yet selected,
       // then add the query's suggest option to the newSelectedOptions to be added to selectedOptions.
       // Otherwise, append the query to the invalid queries string
       if (suggestions && suggestions.length > 0
-          && selectedOption && !selectedOptions.find(item => item.value === selectedOption.value)) {
-        newSelectedOptions.push(selectedOption);
+          && selectedOption && !selectedOptions.find(option => option.value === selectedOption.value)) {
+        if(selectedOption.type === 'mutation'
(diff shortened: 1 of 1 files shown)


chaoran-chen (changes requested):
Hi @ningxie1991, I only took a very brief look at the preview and it looks really good!

I just found one issue: it shouldn't be possible to insert two pangolin lineages (e.g., `B.1.1.7,B.1.617`). Maybe only accept the first one?

image

ningxie1991:
> Hi @ningxie1991, I only took a very brief look at the preview and it looks really good!
> 
> I just found one issue: it shouldn't be possible to insert two pangolin lineages (e.g., `B.1.1.7,B.1.617`). Maybe only accept the first one?
> 
> image

Good catch! When user enters one query then comma one by one this won't happen as the code checks for the max one pangolin lineage rule and won't allow user to enter more than one. But when user just pastes a pre-written comma separated list with more than one pangolin lineages, this happens. I will check and fix that.

ningxie1991:
> > Hi @ningxie1991, I only took a very brief look at the preview and it looks really good!
> > I just found one issue: it shouldn't be possible to insert two pangolin lineages (e.g., `B.1.1.7,B.1.617`). Maybe only accept the first one?
> > image
> 
> Good catch! When user enters one query then comma one by one this won't happen as the code checks for the max one pangolin lineage rule and won't allow user to enter more than one. But when user just pastes a pre-written comma separated list with more than one pangolin lineages, this happens. I will check and fix that.

This is fixed now. Only the first pangolin lineage will be accepted. @chaoran-chen

chaoran-chen (approved):
Looks good to me! (Please just don't forget to fix the formatting)

src/components/VariantSearch/VariantSearch.tsx
```diff
@@ -124,50 +130,146 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
...
+    setInputValue(invalidQueries);
+    // leave options menu open if there are invalid queries from the list
+    // this will show the "No Options" in the menu
+    setMenuIsOpen(invalidQueries !== '');
+  }
+
+  const promiseOptions = (inputValue : string) => {
+    // resets the options to default (where input value is '') when menu is closed
+    if(!menuIsOpen){
+      inputValue = ''
```
spacephoton: Why not `setInputValue('')`?

Should we use `undefined` by default when no input value? (I don't know)

src/components/VariantSearch/VariantSearch.tsx
```diff
@@ -124,50 +130,146 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
...
+  const handleCommaSeparatedInput = (inputValue: string) => {
+    const inputValues = inputValue.split(",")
+    let newSelectedOptions: SearchOption[] = []
+    let invalidQueries = ''
+    for (let query of inputValues) {
+      // remove spaces from both ends of the query
+      query = query.trim()
+      const suggestions = suggestOptions(query);
+      const selectedOption = suggestions.find(option => option.value.toUpperCase() === query.toUpperCase());
+      // If query gets back valid a valid suggest option which is not yet selected,
```
spacephoton: Might be useful to have high level explanation comment on top of the function, rather than details inserted throughout. But this is just a styling thing, so if you prefer like this, no problem.

spacephoton (approved):
Seems good!

To fix format just run the npm script
`npm run format`

src/components/VariantSearch/VariantSearch.tsx
ningxie1991: `setInputValue('')` is called right after at line 205. But because the state change is async and doesn't happen immediately so I still passed the inputValue at line 208. This code is simulating the action that when an option is selected the option menu gets closed and the suggest options get reloaded (it does not happen automatically for selecting from using comma-separated list because onChange is not fired). 

https://github.com/cevo-public/cov-spectrum-website/blob/8944d23088d2d9ce0a803f4003ea52d05d54092e/src/components/VariantSearch/VariantSearch.tsx#L201-L208

The reason why I used `''` instead of `undefined` is because  line 262, the defaultOptions come from '' input. 

https://github.com/cevo-public/cov-spectrum-website/blob/8944d23088d2d9ce0a803f4003ea52d05d54092e/src/components/VariantSearch/VariantSearch.tsx#L262

src/components/VariantSearch/VariantSearch.tsx
ningxie1991: Good suggestions! I added function doc and removed the inline comments.

ningxie1991:
> Looks good to me! (Please just don't forget to fix the formatting)

Formatting is fixed with `npm run format`

ningxie1991:
This is the one warning failing the `Check formatting`, which I am not sure if we can do anything about it.

image

src/components/VariantSearch/VariantSearch.tsx
```diff
@@ -150,18 +256,28 @@ export const VariantSearch = ({ onVariantSelect }: Props) => {
...
-          loadOptions={(query: string) => Promise.resolve(suggestOptions(query))}
+          loadOptions={promiseOptions}
           onChange={(_, change) => {
             if (change.action === 'select-option') {
               setSelectedOptions([...selectedOptions, change.option]);
+              setInputValue('');
+              setMenuIsOpen(false);
             } else if (change.action === 'remove-value' || change.action === 'pop-value') {
               setSelectedOptions(selectedOptions.filter(o => o.value !== change.removedValue.value));
+            } else if (change.action === 'clear') {
```
ningxie1991, line 271: Fixed issue #216

PR #220, opened 2021-08-06
Upgrade zod version to 3.7.1

### LINKED ISSUE #79: Replace Zod
It turns out that Zod is actually [very slow](https://github.com/moltar/typescript-runtime-type-benchmarks). We should switch to a different validation library. Probably something based on `ajv` which is robust and widely used (it uses codegen, so it's fundamentally faster).

After we switch we should re-enable validation for `getNewSamples` and `getInterestingVariants`, since they were disabled in #78 because Zod was too slow to use.

ningxie1991:
Fix for #79

1. Upgraded zod version to 3.7.1
2. Enabled zod parsing in waste water data loading and 3 api requests

@chaoran-chen @TKGZ 
After some testing on my local development environment with zod enabled in waste water data loading and the 3 api request (`/resource/case?`, `/resource/sample2`, `/computed/find-interesting-variants?`), I did not see much performance difference between my local and production. But maybe I am not aware of all the use cases that could bring a huge response load from those api requests. Could you please also check the performance on preview? Thanks!


spacephoton:
Other than package lock, everthing seems to work! :)

ningxie1991:
> Other than package lock, everthing seems to work! :)

I think the current package-lock.json was generated with npm version < 7 so that the `lockfileVersion` = 1 and there are no `packages` section:

https://github.com/cevo-public/cov-spectrum-website/blob/a714299c3789f11b2a60b467d2e897fe61f6696e/package-lock.json#L4-L7

I think my npm version is > 7 so that when I generated the packag-lock.json file, it shows `lockfileVersion` = 2 and includes the `packages` section and all the `"node_modules/` sections (the extra 20000 lines). I googled around and npm verion > 7 apparently generates a new format for the lock file ([stackoverflow thread](https://stackoverflow.com/questions/66150813/is-lockfileversion-2-in-package-lock-json-from-npm-7-compatible-with-older-ve)), which is backwards compatible with npm 6 users according to the answers in the thread. Should I change my npm version to < 7 and generate the lock file again?

chaoran-chen (approved):
@ningxie1991 I think that's working fine! Thank you.

I'm also fine with using the new npm version.

PR #221, opened 2021-08-09
Use React Query for data fetching and error handling

### LINKED ISSUE #56: Proper error handling
- Treat non-200 status from the server as an error
- Show the user something useful when stuff fails to load
- Add [error boundaries](https://reactjs.org/docs/error-boundaries.html)
  - Currently our application will fully crash if the server replies with invalid data (eg. because it's down)

ningxie1991:
For issue #56.

This is to use React Query to catch errors in data fetching and display the error on the UI without crashing the app. 

@chaoran-chen @TKGZ 
As discussed in issue #56, Error Boundary is not applicable for functional components using the `useEffect` hook. And @TKGZ suggested looking into [React Query](https://react-query.tanstack.com/). I checked some blog posts and the documentations and examples. Basically we can replace the whole saving response data in a local state to using `useQuery` which also catches errors and returns the error object in the result. 

Error handling has been implemented for the following api requests:

`getCaseCounts`

image

`getSamples` 

image

`getCaseCounts` + `getSequenceCounts`

image

`getPangolinLineages`

image

getPangolinLineageAliases

https://github.com/cevo-public/cov-spectrum-website/blob/5f99d478ec7339a3f565217ac5ff8e3e6df0b507/src/services/PangolinLineageAliasResolverService.ts#L11-L21


src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -87,26 +88,23 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
   show,
   handleClose,
 }: Props) => {
-  const [caseCounts, setCaseCounts] = useState<CaseCountEntry[] | undefined>();
+  //const [caseCounts, setCaseCounts] = useState<CaseCountEntry[] | undefined>();
```
ningxie1991: We can avoid saving the response data in a local state (see [blog post](https://blog.bitsrc.io/how-to-start-using-react-query-4869e3d5680d) )

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -87,26 +88,23 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
...
   const [showSwissRegions, setShowSwissRegions] = useState(true);
-
   const { dateFrom, dateTo, country } = variantSampleSet.sampleSelector;
-  useEffect(() => {
-    let isSubscribed = true;
```
ningxie1991, line 95: I am not sure why we initialize `isSubscribed` to true here but the variable is to be checked before `setCaseCounts` at the deleted line 100, which will always be true.

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -87,26 +88,23 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
...
+    const data = await getCaseCounts({ dateFrom, dateTo, country }, true, new AbortController().signal);
+    return data;
+  };
 
-    return () => {
-      isSubscribed = false;
-      controller.abort();
-    };
-  }, [dateFrom, dateTo, country]);
+  const { isLoading, isSuccess, error, isError, data: caseCounts, refetch } = useQuery<
```
ningxie1991: We can use these properties returned from `useQuery` to check if the request `isLoading`, `isSuccess` or `isError`, which catches any error in the request.

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -151,8 +149,18 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
           Show cantons
         </span>
       </div>
-      {(!cantonData || !regionData) && <Loader />}
-      {cantonData && regionData && (
+
+      {isLoading && <Loader />}
+      {isError && error && (
```
ningxie1991: We can simply use `isError` and `error` to display the error messages when there's an error in the fetching of data.

src/pages/FocusPage.tsx
```diff
@@ -315,12 +318,14 @@ export const FocusPage = ({
         />
       )}
       {showEstimatedCasesDivGrid && country === 'Switzerland' && (
-        <SwitzerlandEstimatedCasesDivisionModal
-          variantSampleSet={variantSampleSet}
-          wholeSampleSet={wholeSampleSet}
-          show={showEstimatedCasesDivGrid}
-          handleClose={() => setShowEstimatedCasesDivGrid(false)}
-        />
+        <QueryClientProvider client={queryClient}>
```
ningxie1991: We have to enclose the functional component with this `<QueryClientProvider>` tag which I think can be placed at a even higher level like on the whole page component or App component

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -87,26 +88,23 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
...
-      controller.abort();
-    };
-  }, [dateFrom, dateTo, country]);
+  const { isLoading, isSuccess, error, isError, data: caseCounts, refetch } = useQuery<
+    CaseCountEntry[],
+    Error
+  >('caseCounts', fetchCaseCounts);
+
+  useEffect(() => {
+    refetch();
```
ningxie1991: We can call the `refetch` function which is returned from the `useQuery` in the useEffect hook whenever `dateFrom`, `dateTo` and `country` is updated.

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -87,26 +88,23 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
...
-        setCaseCounts(caseCounts);
-      }
-    });
+  const fetchCaseCounts = async () => {
+    const data = await getCaseCounts({ dateFrom, dateTo, country }, true, new AbortController().signal);
+    return data;
+  };
 
-    return () => {
-      isSubscribed = false;
```
ningxie1991, line 106: @chaoran-chen @TKGZ What is the purpose of `isSubscribed` ?

spacephoton: This prevents duplicate fetches, for example if a new fetch is called before the previous one finishes.

spacephoton (approved):
Seems good!

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
chaoran-chen: @ningxie1991, we have some queries that can take quite some time. Do you know what will now happen if the user navigates to another page before the data for the first one was fully loaded? I.e. if the first query is finished after the second query?

ningxie1991: I updated the code so that react query behaves the same. It will cancel request if we close or click out of the modal. And `refetch` data in `useEffect` if `isFetching` = false 

image

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -88,23 +92,25 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
...
-  const { isLoading, isSuccess, error, isError, data: caseCounts, refetch } = useQuery<
+  const { isLoading, isSuccess, error, isError, data: caseCounts, refetch, isFetching } = useQuery<
     CaseCountEntry[],
     Error
-  >('caseCounts', fetchCaseCounts);
+  >('caseCounts', () => {
+    const controller = new AbortController();
+    const signal = controller.signal;
+    const promise = getCaseCounts({ dateFrom, dateTo, country }, true, signal);
+    (promise as PromiseWithCancel<CaseCountEntry[]>).cancel = () => controller.abort();
```
ningxie1991, line 101: This part cancels the query when unmounted. (See [documentation](https://react-query.tanstack.com/guides/query-cancellation))

src/components/SwitzerlandEstimatedCasesDivisionModal.tsx
```diff
@@ -88,23 +92,25 @@ export const SwitzerlandEstimatedCasesDivisionModal = ({
...
+    return promise;
+  });
 
   useEffect(() => {
-    refetch();
-  }, [dateFrom, dateTo, country, refetch]);
+    if (!isFetching) {
+      refetch();
+    }
```
ningxie1991, line 108: This `refetch()` is invoked only when `isFetching` is false inside the `useEffect` hook

chaoran-chen:
By the way, feel free to merge once you have a clean and working version! I'd love to use React Query for future components.

src/components/SampleTable.tsx
```diff
@@ -58,17 +61,14 @@ interface Props {
...
-  const [samples, setSamples] = useState<Sample[] | undefined>(undefined);
-  const [totalNumber, setTotalNumber] = useState<number | undefined>(undefined);
-
-  useEffect(() => {
-    let isSubscribed = true;
+  const mutationsString = variant.mutations.join(',');
+  const { isLoading, isSuccess, error, isError, data: samples, refetch, isFetching } = useQuery<
+    SampleResultList,
+    Error
+  >('samples', () => {
```
chaoran-chen: Hi @ningxie1991. I just started to read about React query. Am I understanding it correctly that the query key should always contain all variables, i.e., in this case the pangolin lineage, the mutations string, the match percentage, the country, and the sampling strategy?

https://react-query.tanstack.com/guides/query-keys

ningxie1991: The query key is an identifier for the particular api request. The parameters for the api request are given in the query function right below:

https://github.com/cevo-public/cov-spectrum-website/blob/f82638d8c3ee4093898dac5bbfd57fd8535ee1ee/src/components/SampleTable.tsx#L72-L78

ningxie1991: Oh I see what you mean. We should add the dependent variables to the query key as well.

ningxie1991:
> By the way, feel free to merge once you have a clean and working version! I'd love to use React Query for future components.

I merged this pull request.

PR #222, opened 2021-08-13
Fix estimated cases plot date range filter bug

### LINKED ISSUE #126: Estimated cases plot: respect the date filter
The sequencing intensity dataset and plot currently are currently not filtered by the date filter. But the dataset is now passed into the estimated case numbers plot which at the same time uses the variantSampleSet which follows the data filter. This mismatch leads to wrong results (zeros) when a date range with a closed end (e.g., 2020) is selected.

Example: 

image

I'm not sure what the best way to resolve it would be...

ningxie1991:
Fix for #126
I replaced the code using `Utils.trimStartBy` and `Utils.trimEndBy` with using `filter`

image


spacephoton:
Looks good!

spacephoton (approved):

chaoran-chen (approved):
Nice. Thanks!

PR #229, opened 2021-09-20
More countries

spacephoton:
Added countries maps:

x Brazil
x France
x Italy
x Japan
x Spain
x China (may have to rename cantons later if more cases)


chaoran-chen (approved):
Cool!

PR #231, opened 2021-09-21
Decouple the KnownVariantList from the Date Range Selector

### LINKED ISSUE #98: "Known variants" plots broken for 2020
When setting the time filter to 2020, the small plots on the "Known variants" buttons don't show any data ("0.0%"). I guess this is because we are filtering by 2020, but also by the last 3 or so months. Either we can make these plots work correctly (fetch the last 3 months of 2020), or just hide them properly (don't show "0.0%").

ningxie1991:
Fix for #98

1. Moved wholeSampleSet data loading from `ExploreFocusSplit` to the `KnownVariantList`
2. Set the dateRange to be `Past3M` across all API calls within `KnownVariantList` so it's not defined everywhere

image

image

### DIFF
--- a/src/components/KnownVariantsList/KnownVariantsList.tsx
+++ b/src/components/KnownVariantsList/KnownVariantsList.tsx
@@ -86,11 +86,6 @@ function selectPreviewVariants(
  * of weeks, we need to fetch slightly more here to ensure we have enough. Therefore,
  * we use past 3 months (Past3M) as the date range across all API calls within KnownVariantsList,
  * e.g. useWholeSampleSet, getPangolinLineages, loadKnownVariantSampleSets.
- * @param country
- * @param samplingStrategy
- * @param onVariantSelect
- * @param selection
- * @constructor
  */
 export const KnownVariantsList = ({ country, samplingStrategy, onVariantSelect, selection }: Props) => {
   const [selectedVariantList, setSelectedVariantList] = useState(VARIANT_LISTS[0].name);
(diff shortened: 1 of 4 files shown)


src/components/KnownVariantsList/KnownVariantsList.tsx
```diff
@@ -74,17 +80,31 @@ function selectPreviewVariants(
...
-  selection,
-  wholeSampleSetState,
-}: Props) => {
+/**
+ * For KnownVariantsList, we don't need very old data, since convertKnownVariantChartData (in load-data.ts)
+ * will only take the latest 2 months. However since our data collection lags by a couple
+ * of weeks, we need to fetch slightly more here to ensure we have enough. Therefore,
+ * we use past 3 months (Past3M) as the date range across all API calls within KnownVariantsList,
+ * e.g. useWholeSampleSet, getPangolinLineages, loadKnownVariantSampleSets.
+ * @param country
```
spacephoton: Don't think we need to add these in the comment, unless you use it with the IDE somehow? (But I think TS infers this stuff anyway)

spacephoton (approved):

src/components/KnownVariantsList/KnownVariantsList.tsx
spacephoton: The @param stuff

ningxie1991: > The @param stuff

Ah they are auto generated when I added the JSDoc by typing `/** enter` . I can remove them.
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

PR #242, opened 2021-09-28
Fix Pangolin Lineage Query Loading Forever Issue

### LINKED ISSUE #237: Pango query often loads forever
I've noticed that the pango query (top panel) often loads forever. Looks like a bug to me.

It's not the case for all queries, just for some, like this one for example: https://cov-spectrum.ethz.ch/explore/Singapore/AllSamples/Past3M/variants/json=%7B%22variant%22%3A%7B%22name%22%3A%22B.1.617.2*%22%2C%22mutations%22%3A[%22s%3A1264L%22]%7D%2C%22matchPercentage%22%3A1%7D

image

ningxie1991:
Fix for #237

https://user-images.githubusercontent.com/3387698/135060872-f20b59ab-8e34-44d2-9eb8-913a7e129738.mp4

### DIFF
--- a/src/components/VariantLineages.tsx
+++ b/src/components/VariantLineages.tsx
@@ -54,11 +54,11 @@ export const VariantLineages = ({
     }[]
   >([]);
   const { dateFrom, dateTo } = dateRangeToDates(dateRange);
-  const unifiedDateFrom = dateFrom && dayjs(dateFrom).format('YYYY-MM-DD');
-  const unifiedDateTo = dateTo && dayjs(dateTo).format('YYYY-MM-DD');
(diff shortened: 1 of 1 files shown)


spacephoton (approved):
Great!

chaoran-chen (commented):
Hi. Great that it works!

Just a word about the variable naming: With `UnifiedDay`, we refer to the classes in https://github.com/cevo-public/cov-spectrum-website/blob/develop/src/helpers/date-cache.ts#L16. In this case, it would be better to just write e.g. `dateFromString`.

chaoran-chen (approved):
Thanks!

PR #243, opened 2021-09-28
UI Loader Fixes 

spacephoton:
* New loading for variant list, replaced because items kept moving (different heights) while loading.
CleanShot 2021-09-28 at 16 12 07
* Move search below list to make it less overwhelming for user (power users who search still can access it with same ease, but I realised that putting so much text and words like 'pangolin' was confusing for users. 
* Prevent loaders from showing at same time as content


ningxie1991 (commented):
Looks good to me :)

ningxie1991 (approved):

chaoran-chen (changes requested):
Hmm, the loader is not working for me. I can only see a white space until the plots eventually appear. I tried Chrome and Firefox on Windows and Chrome and Safari on Mac.

image

Further, I would prefer to have the search on the top of the panel because I think that it is import that people see right away that it is possible to search for lineages or mutations.

corneliusroemer:
I agree, the query bar should be as high up as possible. Definitely not any less discoverable as in current production. Ideally it would be right at the top next to country. What do you think @chaoran-chen ?

ningxie1991:
Ohhh I missed the changes to the position of the variant search bar. Hmmm yeah I agree that it should be more visible on top.

corneliusroemer:
I would prefer it was right at the very top, pinned in the header. But not sure if others agree.

spacephoton:
Sounds good @corneliusroemer @ningxie1991 , will put the bar back on top

@chaoran-chen Will see what the problem is, works locally though

corneliusroemer:
My proposal would be to put it up even higher, you know what I mean?
Putting it at the very very top, where country is. What do you think of
that?

chaoran-chen:
> My proposal would be to put it up even higher, you know what I mean?
Putting it at the very very top, where country is. What do you think of
that?

This wouldn't fit very well with the overall structure in my opinion. The filters on the very top should apply to everything below it which is also the reason why we are moving the date filter to the right panel in #235. The variant preview plots and sequencing intensity plot obviously don't fall under the variant filter. Further, I like to have the variant selection collected on the left side. Also, on mobile, the top filters are not all very obvious
image

spacephoton:
@chaoran-chen @ningxie1991 can one of you check when running locally if the loader appears? I have it appear locally but not when vercel deploys it... :)

PR #244, opened 2021-09-28
Get Github bot to open country update issue twice a month

corneliusroemer:


chaoran-chen:
Hi @corneliusroemer. I have a better idea! LAPIS [1] always has all the countries from the underlying data source (e.g., GISAID).  CoV-Spectrum [2] maintains its own list of country names because it needs to map country names from other sources (e.g., OWID). Could you maybe write a script that checks if a country in [1] is missing in [2] and opens an issue if that happens?

[1] https://cov-spectrum.ethz.ch/gisaid/api/v1/sample/aggregated?fields=country
[2] https://cov-spectrum.ethz.ch/api/v2/resource/country

PR #259, opened 2021-10-15
Removed look behind assertion in regular expression to be compatible with Safari

### LINKED ISSUE #258: Page not loading in Safari - invalid date range regex
The whole website does not work in (at least some versions of) Safari. It just shows a blank page. I tracked down the problem to the following code block:

image

image

(Thanks @laduplessis for reporting!)

ningxie1991:
Fix #258


chaoran-chen (approved):
Thanks!

Issue #250, opened 2021-10-07
Sampling strategy data from GISAID


chaoran-chen:
The current sampling strategy/data type filter in the header is only available for Switzerland and based on our own definition of what "surveillance" is. Earlier this year, this was the only option, because the GISAID dataset did not provide any information about the sampling strategy.

Now, they do. Here are the available sampling strategies:

|sampling strategy|count|
|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
|_null_ | 3262884 |
| Baseline surveillance                                                                                                                                                                                                                                                                                                   | 547196  |
| Random sampling                                                                                                                                                                                                                                                                                                         | 80774   |
| Suspect sampling                                                                                                                                                                                                                                                                                                        | 54822   |
| Baseline surveillance (random sampling)                                                                                                                                                                                                                                                                                 | 29164   |
| unknown                                                                                                                                                                                                                                                                                                                 | 28518   |
| Surveillance testing                                                                                                                                                                                                                                                                                                    | 22857   |
| Baseline Surveillance                                                                                                                                                                                                                                                                                                   | 19136   |
| Non-sentinel-surveillance (hospital)                                                                                                                                                                                                                                                                                    | 5270    |
| Random                                                                                                                                                                                                                                                                                                                  | 4179    |
| Sentinel surveillance (SARI)                                                                                                                                                                                                                                                                                            | 4022    |
| Active surveillance                                                                                                                                                                                                                                                                                                     | 3007    |
| baseline surveillance                                                                                                                                                                                                                                                                                                   | 2824    |
| Surveillance                                                                                                                                                                                                                                                                                                            | 2744    |
| surveillance                                                                                                                                                                                                                                                                                                            | 2499    |
| Baseline Surveilliance                                                                                                                                                                                                                                                                                                  | 2066    |
| Sentinel surveillance (ILI)                                                                                                                                                                                                                                                                                             | 1996    |
| Routine Surveillance                                                                                                                                                                                                                                                                                                    | 1938    |
| S gene dropout                                                                                                                                                                                                                                                                                                          | 1885    |
| Sentinel surveillance                                                                                                                                                                                                                                                                                                   | 1825    |
| Targeted Efforts                                                                                                                                                                                                                                                                                                        | 1450    |
| Vaccination                                                                                                                                                                                                                                                                                                             | 1409    |
[...]

PR #264, opened 2021-10-18
LAPIS and a lot of refactoring

### LINKED ISSUE #32: Describe format of mutations in FAQ
A description of the mutation naming (e.g., ORF8:Q27*) should be added to the FAQ. It should probably contain a link to Nextstrain or Nextclade, explain clearly how the stop codon and deletions are coded and which genes exist.

### LINKED ISSUE #89: Code clean up
I think that we have now quite some unused code (for example in api-types.ts) and it would be great to clean them up :)

### LINKED ISSUE #106: Align date axes
At least whenever a fixed date range is provided (i.e., not "All times" is selected in the date filter in the header), ~all the plots that have a temporal x-axis~ the "sequences over time" and "estimated cases" plots should be aligned.

A few discussion points:

- When 2020 is selected, should the plots start at 01.01, or at the date when the first sars-cov-2 sample was sequenced (in that country), or when the variant was first found (in that country)?
- When 2021 is selected, should the plot go until 31.12.2021 or until today?
- It would be great to have the plots go until today rather than the last (non-zero) data point. However, the default selection (the blue bar) in the current time plot would then always be zero since we won't have any sequences in the ongoing week.

### LINKED ISSUE #118: Pass comparison countries as parameters for the international comparison
Embed comparison with pre-selected countries, for example Switzerland compared to italy and turkey.
image

### LINKED ISSUE #127: Improve code of the estimated case numbers plot
Address comments in #125

### LINKED ISSUE #130: Partially outdated datasets when the server receives new data
Data are not loaded at the same time. Not only is the data fetching process not atomic, but some datasets (e.g., `wholeSampleSet`) are also fetched much less frequently than others (e.g., `variantSampleSet`).

Now that we update the data multiple times a day rather than twice a week, this is becoming an issue. When the data used by the application are not in sync, the shown values can be very wrong!

To fix it, a new endpoint was implemented in https://github.com/cevo-public/cov-spectrum-server/issues/23 ([documentation](https://github.com/cevo-public/cov-spectrum-docs/blob/develop/API.md#data-status)) that returns the timestamp of the last data update **in UTC time**. Two suggestions:

- Whenever the user performs an action, the web app checks the data update timestamp and reloads the page (or only the data) when the data were updated.
- Every minute, the web app checks the data update timestamp. When the data were updated, a non-removable overlay should be shown telling the user that we have new data and it should show a big refresh button.

The embed widgets are partially affected as well. Because the data are not fetched atomically, in rare cases (well, with sufficient visitors, it wouldn't be rare in absolute numbers..), it could also have loaded non-matching datasets. Could we implement that the embed widgets call the data-status endpoint after everything else is fetched and check?

**We can also first fix the first part (ignoring the widgets) as it is more likely! But the widgets will also need a quick fix because we should avoid showing wrong plots by any means!**

**That's just my first idea. Better suggestions are of course highly welcome!**

**In the near future, we should redesign the API so that every response contains the timestamp!**

### LINKED ISSUE #166: [Feature request] Add time of last data update
Outbreak.info shows when the data was last updated. It'd be great to have the same feature on this website.
image

### LINKED ISSUE #183: Make the division sub-plots (of the sequences over time plot) sharable
The division sub-plots are currently not sharable because the sample selector (and the API) do not allow filtering by divisions. The (current/initial) implementation filters the data but passes the original selector.

https://github.com/cevo-public/cov-spectrum-website/commit/8a86592d2b396ea452b64a2036fd3a92253d210f#diff-794371d78f8ad734a009671ae49012f758b400899b3a79c4c44790bcea18ba4cR73-R83

---

In other words: Bring back the "Export" button!

image

### LINKED ISSUE #187: Show original base of the nucleotide mutations
I.e., instead of writing `1234G` as of now, we should write `C1234G`.

### LINKED ISSUE #199: Alias translation is case sensitive
While generally lower case in search input is interpreted as upper case, this is *not* the case for pango lineages, that need to be translated to an alias.

Entering `b.1.1.28.1` yields "no result", while `B.1.1.28.1` gives results. That's because `B.1.1.28.1` = `P.1`

image

image

Lower case yields no match: https://cov-spectrum.ethz.ch/explore/Switzerland/AllSamples/AllTimes/variants/json=%7B%22variant%22%3A%7B%22name%22%3A%22b.1.1.28.1%22%2C%22mutations%22%3A%5B%5D%7D%2C%22matchPercentage%22%3A1%7D

But turning the b from lower to upper case yields results: https://cov-spectrum.ethz.ch/explore/Switzerland/AllSamples/AllTimes/variants/json=%7B%22variant%22%3A%7B%22name%22%3A%22B.1.1.28.1%22%2C%22mutations%22%3A%5B%5D%7D%2C%22matchPercentage%22%3A1%7D

**Proposed fix**
Turn all input in the search box automatically to upper case in javascript. Alternatively, do the same in the backend when parsing URL query parameters. Should be a one-liner.

### LINKED ISSUE #206: It is not possible to filter for wild type amino acids, e.g. S:N501N
It seems to be impossible to exclude mutations from the search.

While I can specify `ORF1a:A2529V` to get me all strains that have V in that position, I *can't* select strains that have A in that place.

Am I doing something wrong? Also, why do you insist on having to specify the amino acid that was there in the wild type? This serves no real purpose. `ORF1a:2529V` would be unique already, adding the `A` in front of the position doesn't add anything, just makes me have to remember more stuff.

image

### LINKED ISSUE #233: [Feature request] Get rid of arbitrary limit of # of countries in `International comparison` plot
There seems to be a limit in the number of countries that can be displayed concurrently in the `International comparison` view.

Is there a reason for this? Why is it (only) 5? Shouldn't it be something like 10 or let the user decide?

image

### LINKED ISSUE #238: Preview plots (and other widgets?) reload after re-focus
The preview plots (and maybe other widgets using react-query) reload after each re-focus.

**Steps to reproduce:**

1. Open CoV-Spectrum
2. Switch to another tab or window
3. Switch back to CoV-Spectrum

**Reason:**

I believe that the issue is caused by the default cache settings of React Query. See: https://react-query.tanstack.com/guides/important-defaults

### LINKED ISSUE #241: [Feature request] Make URL parameters less escape-sequence-heavy
As a power user, I sometimes like to write my queries directly in the URL bar to bypass frontend hassles.

However, I've noticed that the URL queries are very escape sequence heavy, as follows: `/json=%7B"variant"%3A%7B"mutations"%3A["S%3A222V"]%7D%2C"matchPercentage"%3A1%7D`

It'd be nice if the queries didn't contain escape characters. After all, you're not accepting any sort of strings but are passing well-defined parameters. So you can write your own proper encoding. Or probably better, there are libraries for this. Have a look at what Auspice uses for example, much cleaner and nicer on the eye to understand what's going on:
`https://nextstrain.org/groups/neherlab/ncov/europe/2021-08-30?branchLabel=aa&c=gt-S_123,1&l=unrooted&m=div`

Here's how you could rewrite your query parameters (proposal):
`/json=variant&mutations=S_A222V&matchPercentage=1`

Isn't this much cooler than the above mess? I like to see what the query is in the URL bar and you're just making my life hard in this respect ;) 

I think the root cause is that you pass a single string as query parameter, rather than multiple query parameters. It's a shortcut that may have been good in the early days to save time, but for a maintained web app it's a bit, hmm, ugly :)

### LINKED ISSUE #248: [Feature request] Enable country plots, similar to region plots
I'm wondering why there are no country plots, similar to these region plots. The regions are cool, but the countries would be even more useful! Yes, there's the international comparison panel, but it's limited to 6 countries and it's all in one, vs. lots of plots at one glance. Adding country plots just the same way you're doing regions would really be a new cool power feature. And it shouldn't require almost any coding! I'd love the feature in any case ;)

image

### LINKED ISSUE #256: Use LAPIS
Sample information should be fetched from an instance of [LAPIS](https://github.com/cevo-public/LAPIS) instead of the current `/sample2` endpoint.

### LINKED ISSUE #262: Nucleotide mutation search
It's time to support nucleotide mutation searches (currently, we can only search for amino acid mutations).

chaoran-chen:
This PR brings CoV-Spectrum and LAPIS together and refactors almost everything that is invisible (i.e., not related to the view). There are two major architectural changes:

- Data selection, fetching and transformation: We mostly use two data frames: sequencing data and case data. We filter and aggregate the data in different ways.
    - **Before:** The data were fetched and aggregated in very different places. For example, the time distribution plot requested the whole, quite detailed sample set and then performed the aggregation. The embed widget then also fetched the whole dataset rather than getting only the time distribution data. Further, the variant selector was 
    - **Now:** The data fetching and typical data transformation classes and functions are now in the `data` folder. There is a new interface called `Dataset`. A `Dataset` contains a `Selector` and a `Payload`. The selector is a set of filters and clearly defines the dataset. It is possible to fetch the payload of a dataset from the API given its selector. It is possible to transform between datasets: e.g., a `DateCountSampleDataset` can be derived from a `
DetailedSampleAggDataset` by grouping/aggregating. A chart/widget defines a list of datasets that it requires.
- The plot classes
    - **Before:** We had three classes per plot: `...Chart, ...Plot, ...PlotWidget`. The `...Chart` and `...Plot` files did not have the same prefix (e.g., `VariantAgeDistributionPlot` uses `TypeDistributionChart`). A `...Plot` contained a `...Chart` but it wasn't clear which logic belongs to which class. Further, some plots fetched their own data, and others only used data passed through the arguments.
    - **Now:** There are always a `...Chart`and a `...ChartWidget` file. Optionally, there is a `...ChartInner` file that has the same prefix. All the files are in the `widget` folder. The plots do not fetch any own data.

LAPIS and the improvement of the data flow made the fixing of many open tasks very trivial. Below is the full list of changes:

#### Improvements

- Sequences are fetched from [LAPIS](https://github.com/cevo-public/LAPIS) instead of the `/resource/sample2` endpoint (fix #256).
- A new and fully-rewritten version of the CoV-Spectrum server will be used (https://github.com/cevo-public/cov-spectrum-server/pull/25).
- The variant selector in the URL is encoded easier readable using multiple request params instead of encoding everything into a JSON (fix #241).
    - Example URL: `/explore/Switzerland/AllSamples/AllTimes/variants?aaMutations=S%3AN501N&pangoLineage=B.1.1.7`
    - The old URLs are not supported anymore. The user will be directed to the overview page of the country but won't be redirected to the correct page.
- The user can search for nucleotide mutations (fix #262).
- The user can search for sequences where a base was not mutated. For example, `S:N501N` (fix #206)
- In the (common) nucleotide mutation section at the bottom of the focus page, the nucleotide mutations contain the original base (fix #187)
- The user can specify a list of countries that should be shown by default in the international comparison plot (fix #118).
- The user can select any number of countries in the international comparison plot (fix #233)
- If a region or the world is selected, the sub-plots (of the time and age distribution) will show country instead of division plots (fix #248).
- At the bottom of the explore page, the date and time of the last update is displayed (fix #166)
- Fix #106
- Fix #183
- Fix #32

#### Destructive changes

The implementation of a few older features was unclean, and some private endpoints are not supported by LAPIS. This leads to the removal of a few functions:

- The sampling strategy was removed (shall come back with #250).
- The Nextclade integrations were removed. It hasn't been used in quite a while and might not come back.
- The Nextstrain integration was removed (might come back with #272).
- The private Swiss zip code map was removed (might come back with #271).
- The private sample detail table was removed. No one uses it.
- The plots for the Swiss Sentinella regions were removed (shall come back with #273).

#### Minor code improvements

- Fix #89
- Fix #127

#### Bug fixes

- Fix #238
- LAPIS responses contain a `dataVersion` field which is the timestamp of the last data update. Data with different versions may not be used together, for example, to calculate proportions. We now check the data version with every request and reload the page if there is a new version (fix #130)
- Fix #199 

## Remaining TODOS (for this PR)

- [x] Clean up code
- [x] Update `/docs`
- [x] #32
- [x] Fix countries for which we do not have case count data.
- [x] Show country-wise estimated cases sub-plots for regions and "world" (related to #248)
- [x] Fix country mapping between our and GISAID's country names. For example, we write "United States" but GISAID uses "USA".


chaoran-chen:
@ningxie1991, @TKGZ, the code still needs clean-up but it would be great if you could already start testing! (Please also click on the show more buttons :P)

@corneliusroemer, you are of course also very welcome to test it and give feedback!

corneliusroemer:
Can't wait for this to be in production, love the improvements, like country list sorted by first occurrence etc!

chaoran-chen:
country list sorted by first occurrence? Do you mean in the search? I actually forgot to do any sorting, wouldn't an alphabetical sort be the best? Or do you mean another country list?

corneliusroemer:
This list is now sorted by first occurrence, which I like, because I hated searching by eye for the smallest number. Indexed by first occurrence is exactly the kind of search one does in that table (at least I do)
image

chaoran-chen:
I believe that the code is stable enough for merge. @ningxie1991 and @TKGZ, it would be great if you could give it an exhaustive review nonetheless!

corneliusroemer:
Amazing!

Issue #175, opened 2021-06-18
Compare Variants


spacephoton:
How to compare variants?

- Somehow reuse the existing?
- Analyze all aspects? Geo, timeplot, metadata, etc => Big overall picture
- What mutations make two lineages different?
- Identify properties that make mutation important
@chaoran-chen anything else to add?

corneliusroemer:
- Estimate relative transmission advantages using multinomial fits

chaoran-chen:
I see three important aspects:

**(I) How should the user select multiple variants to compare? How does the user experience/journey look like?**

I don't really have a good suggestion, yet. A simple way is probably to introduce a second tab on the explore page (as we had before).

**(II) Should the user be able to select overlapping variants? (I.e., allow variants defined by mutations, and e.g., comparing B.1\* with B.1.1)?**

Yes! We definitely need to support this as it is a core feature of cov-spectrum (also conceptually) to support searches by mutations. For most of the plots, it shouldn't be a problem. "Estimate relative transmission advantages using multinomial fits" as @corneliusroemer suggests, might be difficult and we might not show it for overlapping variants until we find a mathematically sound solution.

**(III) Which plots do we show and how?**

I suggest that we keep the focus page structurally the same as now - as far as possible. It would be great if we can provide the user with the same type of plots and structure.

- Sequences over time: For the comparison, we might want to use a line plot instead.
- Estimated cases: Omit the confidence intervals, otherwise, just add more lines.
- Demographics: Just add more bars (in different colors)?
- Hospitalization probabilities: That's going to be hard to implement, I guess.. Currently, the plot already compares two variants and it looks okay, but more than two might be difficult. We might want to show the probabilities of different variants next to each other.
-  Estimation of the current advantage: Will be replaced by a more suitable model.
- Geography: Extending the table should be easy.
- Wastewater prevalence: Nope! Probably not going to show it on the comparison site.
- International comparison: Nope!
- Publication and pre-Prints: don't show as well
- The mutations: We should show both the shared and not shared mutations, maybe using a table (rows are the mutations, columns are the variants, write the percentages in the cells?)

corneliusroemer:
Can you give examples for what sort of variants you'd like to compare and what the reason for this is?

Is it to do cleaner comparisons, that are more stable between countries, that is, you avoid the fact that if you consider one variant against the rest, the rest may differ a lot between countries.

chaoran-chen:
One current use case I have: compare the different AY variants.

PR #302, opened 2021-10-27
Calculate Jaccard similarity as uniqueness score for mutations

chaoran-chen:
Many thanks to @adrian-lison for the discussion on uniqueness/similarity scores and for the suggestion to use Jaccard!

This feature helps us to find mutations that are unique to a variant.

Here's an example with B.1.1.7:

image

### DIFF
--- a/src/components/KnownVariantsList/KnownVariantsList.tsx
+++ b/src/components/KnownVariantsList/KnownVariantsList.tsx
@@ -183,7 +183,6 @@ export const KnownVariantsList = ({
       return [];
     }
     const variantList = VARIANT_LISTS.filter(({ name }) => name === selectedVariantList)[0];
-    console.log(1);
     return selectPreviewVariants(
       variantList.variants,
       pangoCountDataset.data.getPayload(),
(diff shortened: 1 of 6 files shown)


corneliusroemer:
What's the Jaccard similarity calculated with respect to? All sequences in that timeframe in that country?

corneliusroemer:
This is good! When I sort according to similarity, I get the new mutations that have arisen within AY.4.2 showing up, these are the mutations I'm tracking already, at least there's a good overlap, so this can be very helpful!
image

corneliusroemer:
I love how the similarities pop up one after another async 👯 
image

chaoran-chen:
It's calculating the similarity of the set of sequences belonging to the selected variant and the set of sequences belonging to the mutation. The country and date range filters apply as usual.

corneliusroemer:
Right, so if a homoplasic mutation appears, like N501Y in AY.4.2, it would have a low Jaccard because N501Y has already popped up a lot in other sequences.
In that sense the mutations highlighted may actually be deleterious ones, since the ones giving an advantage will have already shown up frequently in the past.
It may be interesting to make the comparison with lineages one up in the hierarchy, for example AY.4 vs AY.4.2

chaoran-chen:
> It may be interesting to make the comparison with lineages one up in the hierarchy, for example AY.4 vs AY.4.2

I can imagine that it will be eventually possible with #175. Then, you can select the variants that you would like to compare.

The hierarchy concept itself does not really fit cov-spectrum because a "variant" is often not (just) a pango lineage.

PR #312, opened 2021-10-31
Add a line chart as a second visualization for the time distribution

chaoran-chen:
@corneliusroemer, what do you think? :)

image

Todo

- [x] Cutoff when selecting 2021 as date range
- [x] Show y-axis ticks as percent values

### DIFF
--- a/src/components/NamedCard.tsx
+++ b/src/components/NamedCard.tsx
@@ -8,12 +8,19 @@ export enum NamedCardStyle {
   CONFIDENTIAL,
 }
 
+export type TabConfig = {
+  labels: string[];
+  activeTabIndex: number;
+  onNewTabSelect: (tabIndex: number) => void;
+};
+
 interface Props {
   title: string;
   toolbar?: React.ReactChild | React.ReactChild[];
   children: React.ReactChild | React.ReactChild[];
   style?: NamedCardStyle;
   description?: string;
+  tabs?: TabConfig;
 }
 
 export const Card = ({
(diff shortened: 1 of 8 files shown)


spacephoton:
Hmm I'm not sure it's a good ideas as it adds UI complexity to visualize the same data

chaoran-chen:
Do you have a suggestion for improvement? I think that both visualizations have their own value (the line plot is more smoothed and shows the uncertainty; the bar plot shows the exact number of cases per week).

OWID takes a similar approach and, for example, offers a chart, a map and a table for the same data (https://ourworldindata.org/explorers/coronavirus-data-explorer), and I think it's very nice.

spacephoton:
The main thing is that it's very clearly different data types, not lines VS bars. Maybe we can combine the way Google does, or you don't like it? 
CleanShot 2021-10-31 at 15 58 47

corneliusroemer:
This is great! Definitely useful already as is.

I don't understand @TKGZ's criticism, can you elaborate? I also don't understand your Google suggestion.

One thing I'd suggest is maybe using 50% CI, as the current CIs are too broad to be useful. Or if you don't want to get rid of 95% CI why not use both with two different shades of grey/blue? Like this:
image

chaoran-chen:
> One thing I'd suggest is maybe using 50% CI, as the current CIs are too broad to be useful.

I created GenSpectrum/dashboards#63.

corneliusroemer:
Cool, what about the other suggestion, using two colors? That'd be better than customizable IMO. Too complicated?

corneliusroemer:
The date range overflows on my screen causing the whole graph to be pushed down:
image

It'd be cool if the dates wouldn't break to a new line like occasionally here
image

chaoran-chen:
> Cool, what about the other suggestion, using two colors? That'd be better than customizable IMO. Too complicated?

Implementation-wise, it should be doable, but for a default plot, it seems too to me. Also, if we implement such a feature, I would like to apply it to all plots with a CI.

> The date range overflows on my screen causing the whole graph to be pushed down:

Good catch, I'll try to fix it.

corneliusroemer:
I just feel like the 95% CIs are basically useless, far too broad to be meaningful. And so I'd like to see 50% being visible by default. But I see you agree, you like to be on the safe side showing everyone the 95%. A compromise would be to display both ;)

chaoran-chen:
> I just feel like the 95% CIs are basically useless, far too broad to be meaningful. And so I'd like to see 50% being visible by default. But I see you agree, you like to be on the safe side showing everyone the 95%. A compromise would be to display both ;)

Let's deal with this in another PR :)

> > The date range overflows on my screen causing the whole graph to be pushed down:
> 
> Good catch, I'll try to fix it.

I don't actually know how to fix it. But I moved it to a new issue: GenSpectrum/cov-spectrum-website#316

PR #344, opened 2021-11-19
Moved variant header out of loader and changed the datepicker actions

### LINKED ISSUE #305: Variant header and date range picker should stay while content is loading?
When I implemented the date range picker, I moved the variant header and the date picker out of the loadable content, so that they stay when user changes the date and only the content below would load. Now the header and date picker will disappear while loading as well. Should we keep the old way? For me, if the header disappear, it seems like something is going wrong.

@chaoran-chen @TKGZ  What do you think?

image

image

### LINKED ISSUE #337: Date picker already reloads when typing single digit: should only reload on enter
When I manually edit the date range, the page reloads with a new date range, as soon as I edit for example one month.

The whole page turns white and I can't actually finish setting the date range.

This is particularly problematic as covSpectrum has very slow load times. So I have to edit the URL instead of using the date entry field - which is alright for me but not very user friendly.

ningxie1991:
Fix for #305 and #337

1. Moved `VariantHeader` out of `FocusPage` so that we don't reload it when data is reloaded
2. Put variant selector in state and set new value when variant dataset is changed
3. Reload variant header when another variant is selected
4. Date picker only sends new requests when user clicks out (if they select new date from calendar) or hits enter (if they input new date by typing)

@chaoran-chen @TKGZ 

https://user-images.githubusercontent.com/3387698/142632364-3d5beaa6-baa3-49ae-bb02-dd2418ec8621.mp4

### DIFF
--- a/src/components/DateRangePicker.tsx
+++ b/src/components/DateRangePicker.tsx
@@ -17,8 +17,13 @@ const today = new Date();
 
 export const DateRangePicker = ({ dateRangeSelector }: Props) => {
   const { dateFrom, dateTo } = dateRangeSelector.getDateRange();
-  const [startDate, setStartDate] = useState<Date>(dateFrom ? dateFrom.dayjs.toDate() : minimumDate);
-  const [endDate, setEndDate] = useState<Date>(dateTo ? dateTo.dayjs.toDate() : today);
+  const initialStartDate = dateFrom ? dateFrom.dayjs.toDate() : minimumDate;
+  const initialEndDate = dateTo ? dateTo.dayjs.toDate() : today;
+  const prevDateFrom = globalDateCache.getDayUsingDayjs(dayjs(initialStartDate));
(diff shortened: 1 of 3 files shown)


ningxie1991:
@chaoran-chen 
From this I also found another issue, not sure if it is considered a bug. 
When user choses a date range for one variant, then they decide to select another variant, the date range is not reset to AllTimes. Is this expected?

https://user-images.githubusercontent.com/3387698/142634443-d68c5180-6ed4-4897-ac6a-87d973e6f34c.mp4

ningxie1991:
- [x] Found a bug: clicking out without date change should not trigger a request! (fixed)

ningxie1991:
@chaoran-chen @TKGZ Do you have any feedback on this? Can I merge it?

chaoran-chen (approved):
Hi. Sorry for the late reply. It looks good! Thanks.

PR #349, opened 2021-11-22
fix: add wildcards to all editor's choice lineage defs

### LINKED ISSUE #348: All editor's choice lineages should by default be wildcard queries, unless explicitly not so
In order to make it less tedious to maintain editor's choice lineages, the default should be a wild card.

For example, it's harmful, that the editor's choice is `AY.4.2` not `AY.4.2.1` - some relevant `AY.4.2*` are missed this way.

All editor's choices should just start off as wildcards unless there's good reason not to put a wild card there, but only then!

image

Same applies to Alpha, there are now Alpha sublineages, so they should just be included through `B.1.1.7*` etc

corneliusroemer:
resolves #348

### DIFF
--- a/src/components/KnownVariantsList/variantLists.json
+++ b/src/components/KnownVariantsList/variantLists.json
@@ -4,7 +4,7 @@
     "fillUpUntil": 12,
     "variants": [
       {
-        "pangoLineage": "B.1.1.7*,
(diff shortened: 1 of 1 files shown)


chaoran-chen (approved):
Thanks!

PR #362, opened 2021-12-06
Global map total case slider

spacephoton:
Time range slider for global map.

CleanShot 2021-12-06 at 17 53 31

**(Probably) Next steps:**
- Do the same but for variant proportion (another tab)
- Make both types of components sharable
- Include them in the main region pages
- Include other region types


spacephoton:
Added animation button (which runs for 10 seconds through timeline):
CleanShot 2021-12-07 at 15 40 34

chaoran-chen (approved):
Looks cool!

Issue #367, opened 2021-12-12
Layout restructuring: move search bar and simplify ExplorePage


chaoran-chen:
Here's an idea for a small restructuring. The search bar is on the right side. If we implement #366, the sequencing intensity plot is already on the landing page and does not need to appear on the ExploreFocusSplit-view. This allows us to make the left area narrower.

![image](https://user-images.githubusercontent.com/18666552/145729839-c9dcf038-04df-4656-bcda-71e5b3eccc03.png)

ningxie1991:
I like this layout!

chaoran-chen:
Actually, we can even abandon the two-sided layout, no need for two scroll bars anymore.

![image](https://user-images.githubusercontent.com/18666552/146208589-b45f3364-11e4-40e7-8c9d-f56c91e70e15.png)

ningxie1991:
> Actually, we can even abandon the two-sided layout, no need for two scroll bars anymore.
> 
> ![image](https://user-images.githubusercontent.com/18666552/146208589-b45f3364-11e4-40e7-8c9d-f56c91e70e15.png)

Agreed. I was about to suggest the same that we should not have two scroll bars.

spacephoton:
What about keeping it horizontal by default?

ningxie1991:
And what about moving the date range picker up from the variant header to where the search bar is?

chaoran-chen:
Sure, sounds good!

PR #373, opened 2021-12-20
Select variants for comparison

### LINKED ISSUE #309: Experiment the selection method of multiple variants for comparison
Related to #175

### ToDo

1. Add multiple variants to the url
Some options:
    1. `/explore/Switzerland/AllSamples/Past6M/variants?pangoLineage=B.1.1.529*&aaMutations=N%3AR203K&pangoLineage=B.1.1.112&pangoLineage=B.1.617.2*&pangoLineage=N.10&aaMutations=N%3AM1X&pangoLineage=P.1.17.1`
    2. `/explore/Switzerland/AllSamples/Past6M?variants=%5B%7B%22aaMutations%22%3A%5B%22N%3AR203K%22%2C%22N%3AS235F%22%5D%2C%22pangoLineage%22%3A%22B.1.1.529*%22%7D%2C%7B%22pangoLineage%22%3A%22B.1.617.2*%22%7D%5D`
    Option i is most straightforward but not good for parsing different variants with various combination of properties. Option ii is encoding the whole list/array of variants as they are.

    Suggestion:
    We encode the array of variants.
    Current url param encoding implementation only works for one variant. 
    https://github.com/cevo-public/cov-spectrum-website/blob/6093a9d180ec42e4218d8b3870fcfe17aba778f5/src/data/VariantSelector.ts#L22-L35
If we use JSON.stringify, we don't need to manually check if a property is undefined or not. 
```
const variants = [{
  aaMutations: ['N:R203K', 'N:S235F'],
  gisaidClade: undefined,
  nextstrainClade: undefined,
  nucMutations: undefined,
  pangoLineage: "B.1.1.529*",
  variantQuery: undefined
}, {
  aaMutations: [],
  gisaidClade: undefined,
  nextstrainClade: undefined,
  nucMutations: undefined,
  pangoLineage: "B.1.617.2*",
  variantQuery: undefined
}]
console.log(JSON.stringify(variants))

const filteredVariants= JSON.stringify(variants.map(function(selector) {
    if (selector.aaMutations?.length == 0) {
      selector.aaMutations = undefined
    }
    if (selector.nucMutations?.length == 0) {
      selector.nucMutations = undefined
    }
    return selector;
  }));

const params = new URLSearchParams()
params.set("variants", JSON.stringify(filteredVariants))
console.log(params.toString())

const query = new URLSearchParams(params)
console.log(JSON.parse(query.get("variants")))

Output:
> '[{"aaMutations":["N:R203K","N:S235F"],"pangoLineage":"B.1.1.529*"},{"aaMutations":[],"pangoLineage":"B.1.617.2*"}]'
> "variants=%22%5B%7B%5C%22aaMutations%5C%22%3A%5B%5C%22N%3AR203K%5C%22%2C%5C%22N%3AS235F%5C%22%5D%2C%5C%22pangoLineage%5C%22%3A%5C%22B.1.1.529*%5C%22%7D%2C%7B%5C%22pangoLineage%5C%22%3A%5C%22B.1.617.2*%5C%22%7D%5D%22"
> '[{"aaMutations":["N:R203K","N:S235F"],"pangoLineage":"B.1.1.529*"},{"pangoLineage":"B.1.617.2*"}]'
```

2. Keep a list of variants
Currently, the code is written for displaying one Pango or variant only. The implementation now does not work well with multiple Pango or variants. 
Current implementation:
    1. Click on a variant card in the preview / search for a variant 
    2. Calls explore-url.ts script to update the url
    3. Use URLSearchParams to query the url for any variants 
   https://github.com/cevo-public/cov-spectrum-website/blob/6093a9d180ec42e4218d8b3870fcfe17aba778f5/src/helpers/explore-url.ts#L180-L195
    Then we return the found variant back to the ExploreFocusSplit page.
    We need to maintain a list where we can add and remove items. 
    Suggestion:
    We keep the list in  explore-url to set the url query string. We also need to modify the code in ExploreFocusSplit to keep dataset for multiple variants. 
    The current implementation searches for on selector at a time. Shall we create an API for multi-selecors query?
https://github.com/cevo-public/cov-spectrum-website/blob/6093a9d180ec42e4218d8b3870fcfe17aba778f5/src/pages/ExploreFocusSplit.tsx#L49-L61
     We could write a `useQueries` hook to execute parallel queries or another API that takes an array of variant selectors as input.

3. Focus page needs to show data for multiple variants
    Suggestion:
    We can keep the Explore panel functions mostly, just by adding an "Add" button to the search input, so we can use the same search box to add variants to the list. Also, we will add multi select on the preview grid, so user can select multiple mini charts to see comparison.

ningxie1991:
For #309

@chaoran-chen @TKGZ 
This is roughly how it will work:

- The double scroll bars are removed
- Search bar is moved to the right in focus page
- A "Compare Variants" toggle switch is added
- Everything works as it is in Non Compare Mode
- The variant header becomes a list of buttons in compare mode, clicking on a variant button will remove it from the list.
- The KnownVariant list becomes multi selectable in "Compare Mode" (clicking = select, click again = deselect)
- The code infers from the URL whether the website should be in Compare Mode or not. (variant list length > 1 ?  compare mode)

https://user-images.githubusercontent.com/3387698/146802987-10baabb8-71df-4b3b-a6bc-693410aa07c1.mp4

But there are some questions:

1.  Why are we keeping empty array for aaMutations for all the known variants in the json file? As this empty array can interfere with object deep comparison in some places when comparing a variant with the variant list. In this draft pull request, I removed the empty aaMutations list for the first few known variants in the Editor's Choice list, this way, the comparison works for handling multiple variants in a list. 
 https://github.com/cevo-public/cov-spectrum-website/blob/434f87d3117e60d3aa59dcb4cb6e327292be94e4/src/components/KnownVariantsList/variantLists.json#L10-L15

2. Should we restrict the search bar in compare mode to only suggest variants that are not selected? Or we can add a condition to the "Add" logic to not add duplicated variants to the list.

### Screenshot:
image

### ToDo
- [ ] Known variants json --> should keep them empty aaMutations list or not?
- [ ] Fix the focus page to display the actual comparison
- [ ] Improve the search bar logic in Compare Mode
- [ ] UI adjustments of style

### DIFF
--- a/src/components/CompareModeToggle.tsx
+++ b/src/components/CompareModeToggle.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+
+interface Props {
+  handleModeChange: () => void;
+}
+
+export const CompareModeToggleSwitch = ({ handleModeChange }: Props) => {
+  return (
+    <div className='custom-control custom-switch custom-switch-lg'>
+      <input
+        type='checkbox'
+        className='custom-control-input'
+        id='customSwitches'
+        onClick={handleModeChange}
+      />
+      <label className='custom-control-label' htmlFor='customSwitches'>
+        Compare Variants
+      </label>
+    </div>
+  );
+};
--- a/src/components/CompareModeToggle.tsx
+++ b/src/components/CompareModeToggle.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
 
 interface Props {
+  compareMode: boolean;
   handleModeChange: () => void;
 }
 
-export const CompareModeToggleSwitch = ({ handleModeChange }: Props) => {
+export const CompareModeToggleSwitch = ({ compareMode, handleModeChange }: Props) => {
   return (
     <div className='custom-control custom-switch custom-switch-lg'>
       <input
@@ -12,6 +13,7 @@ export const CompareModeToggleSwitch = ({ handleModeChange }: Props) => {
         className='custom-control-input'
         id='customSwitches'
         onClick={handleModeChange}
+        checked={compareMode}
       />
       <label className='custom-control-label' htmlFor='customSwitches'>
         Compare Variants
--- a/src/components/KnownVariantsList/KnownVariantCard.tsx
+++ b/src/components/KnownVariantsList/KnownVariantCard.tsx
@@ -85,7 +85,9 @@ export const KnownVariantCard = ({ name, chartData, recentProportion, onClick, s
   return (
     <Card
       as={StyledCard}
-      className={`shadow-md border-0 m-0.5 hover:border-4 transition delay-20 duration-300 ease-in-out transform hover:scale-105 hover:shadow-xl w-full`}
+      className={`shadow-md m-0.5 hover:border-4 transition delay-20 duration-300 ease-in-out transform hover:scale-105 hover:shadow-xl w-full ${
+        selected ? 'border-4 border-blue-500' : ''
+      }`}
       onClick={onClick}
       selected={selected}
     >
--- a/src/components/KnownVariantsList/KnownVariantsList.tsx
+++ b/src/components/KnownVariantsList/KnownVariantsList.tsx
@@ -20,7 +20,8 @@ const KnownVariantCardLoader = (
     <div className={`h-20 bg-gradient-to-r from-gray-400 to-gray-300 rounded w-full`}></div>
   </div>
 );
-const getLoadVariantCardLoaders = () => {
+
+const getLoadVariantCardLoaders = (isLandingPage: boolean) => {
   let loaders = [];
   for (let i = 0; i < 12; i++) {
     loaders.push(KnownVariantCardLoader);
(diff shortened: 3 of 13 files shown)


chaoran-chen:
Hi @ningxie1991!

> The double scroll bars are removed
Search bar is moved to the right in focus page

Would it be easy to extract this part to a separate PR so that we can merge it quickly? This would close #367.

> 1. Why are we keeping empty array for aaMutations for all the known variants in the json file?

Probably only due to historical reasons. I think that you can remove the empty arrays safely.

> 2. Should we restrict the search bar in compare mode to only suggest variants that are not selected? Or we can add a condition to the "Add" logic to not add duplicated variants to the list.

I like the second suggestion, i.e., preventing adding duplicated variants when clicking on "Add".

ningxie1991:
> Hi @ningxie1991!
> 
> > The double scroll bars are removed
> > Search bar is moved to the right in focus page
> 
> Would it be easy to extract this part to a separate PR so that we can merge it quickly? This would close #367.

**Yes, I will do that.**

> 
> > 1. Why are we keeping empty array for aaMutations for all the known variants in the json file?
> 
> Probably only due to historical reasons. I think that you can remove the empty arrays safely.
> 
> > 2. Should we restrict the search bar in compare mode to only suggest variants that are not selected? Or we can add a condition to the "Add" logic to not add duplicated variants to the list.
> 
> I like the second suggestion, i.e., preventing adding duplicated variants when clicking on "Add".

chaoran-chen:
Took over by #382.

PR #374, opened 2021-12-22
Moved search bar to above the focus page on the right + use one scroll bar only

### LINKED ISSUE #367: Layout restructuring: move search bar and simplify ExplorePage

ningxie1991:
For #367 

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,19 +14,32 @@ import { baseLocation } from './index';
 import StoriesOverview from './stories/StoriesOverview';
 import StoryRouter from './stories/StoryRouter';
 import { defaultDateRange, defaultSamplingStrategy } from './helpers/explore-url';
+import dayjs from 'dayjs';
+import { getCurrentLapisDataVersionDate } from './data/api-lapis';
+import { sequenceDataSource } from './helpers/sequence-data-source';
+import { ExternalLink } from './components/ExternalLink';
+import { VercelSponsorshipLogo } from './components/VercelSponsorshipLogo';
+import styled from 'styled-components';
 
 const isPreview = !!process.env.REACT_APP_IS_VERCEL_DEPLOYMENT;
+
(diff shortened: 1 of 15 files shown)


ningxie1991:
Some ideas to move the data range picker:

image

image

image

chaoran-chen:
Maybe let it depend on the screen width?

spacephoton:
what about the date range picker as part of the header?

chaoran-chen:
Didn't we just move it out of the header a while ago? I remember two keys reasons: 1. the date picker does not apply to the preview plots on the left, 2. the header is already very crowded.

ningxie1991:
Yes, those were the reasons. 

What do we think about these? I thought of a prettier way for the date range picker:

1.
image

2.
image


3. 
image


This react-bootstrap-daterangepicker can potentially contain the predefined ranges like this:

4.
image

5.
image

6.
image

7.
image

chaoran-chen:
Both approaches look interesting! Could you maybe push them somewhere so that I can try them out in the preview?

ningxie1991:
> Both approaches look interesting! Could you maybe push them somewhere so that I can try them out in the preview?

Option 1:
https://cov-spectrum-develop-4w0n2ereq-cov-spectrum.vercel.app/explore/Switzerland/AllSamples/Past6M/variants?pangoLineage=B.1.1.529*

chaoran-chen:
I like the first option! Let's take it.

Could we have the preview plots a bit more narrow? I think that they can remain the same width as before and that we should give the additional space rather to the right side. The "Sequencing intensity" plot can be removed as it is already on the landing page.

PR #403, opened 2022-02-15
Bug/overflow fix

gautier-collab:
Issue #403 : fixed the overflow issues on small screens.

That encompasses:
 - the position of "Export" and "Show regions" toolbar buttons
 - Sequences over time chart
 - Transmission advantage chart
 - International comparison chart

PR #404, opened 2022-02-15
Feature/document title

### LINKED ISSUE #365: Document title
Currently, we only show "covSPECTRUM" as the document title. It would be better to show page-specific titles. This would e.g. make it easier to find the right tab and to create descriptive bookmarks.

I suggest:

- ExplorePage
```
<country> - covSPECTRUM
Ex.: Europe - covSPECTRUM
```

- DeepSequencingCoveragePage
```
Sequencing coverage - <country> - covSPECTRUM
Ex.: Sequencing coverage - Europe - covSPECTRUM
```

- FocusPage and the other Deep-pages
```
<variant display name> - <country> - covSPECTRUM
Ex.: B.1.617.2* + S:N501Y - Europe - covSPECTRUM
```

For now, I suggest that we don't give the Deep-pages specific names because the names would be getting quite long.

- The other pages
```
Stories - covSPECTRUM
Wastewater in Switzerland - Stories - covSPECTRUM
About - covSpectrum
Login - covSPECTRUM
```

gautier-collab:
Page title changes along the navigation, like indicated in the issue #365

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,6 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-puppeteer": "^6.0.0",
     "postcss": "^8.3.0",
-    "prettier": "^2.2.1",
     "puppeteer": "^10.4.0",
     "resize-observer-polyfill": "^1.5.1",
     "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2"
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-puppeteer": "^6.0.0",
     "postcss": "^8.3.0",
-    "prettier": "^2.5.1",
+    "prettier": "^2.2.1",
     "puppeteer": "^10.4.0",
     "resize-observer-polyfill": "^1.5.1",
     "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2"
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-puppeteer": "^6.0.0",
     "postcss": "^8.3.0",
+    "prettier": "^2.2.1",
     "puppeteer": "^10.4.0",
     "resize-observer-polyfill": "^1.5.1",
     "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.2"
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-puppeteer": "^6.0.0",
     "postcss": "^8.3.0",
-    "prettier": "^2.2.1",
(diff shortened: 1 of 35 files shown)


chaoran-chen:
Hi. Could you please fix the formatting (here and in #403)? You just need to run `npm run format`.

src/pages/ExplorePage.tsx
```diff
@@ -42,6 +42,15 @@ export const ExplorePage = ({ isSmallScreen }: Props) => {
     ({ selector }, { signal }) => CaseCountData.fromApi(selector, signal)
   );
 
+  useEffect(() => {
+    // Include the location of interest in the page title
+    let locationObj: any = exploreUrl?.location!
+    let place: String = 'country' in locationObj
```
chaoran-chen: Is there a reason why the type `String` is used instead of `string`? If not, please use `string` (here and at other places).

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean

> The type names String, Number, and Boolean (starting with capital letters) are legal, but refer to some special built-in types that will very rarely appear in your code. Always use string, number, or boolean for types.

src/pages/FocusPage.tsx
```diff
@@ -42,6 +42,18 @@ export const FocusPage = ({ isSmallScreen }: Props) => {
...
   }
+
+  useEffect(() => {
+    // Include the variant name and location of interest in the page title
+    let variantObj = typeof exploreUrl.variants == 'object' ? exploreUrl.variants[0] : {}
+    let variant = 'pangoLineage' in variantObj? variantObj['pangoLineage'] : 'Variant'
+    let locationObj: any = exploreUrl?.location!
+    let place: String = 'country' in locationObj
+      ? locationObj['country'] 
+      : ('region' in locationObj? locationObj['region'] : 'World')
```
chaoran-chen: I think that it would be better to extract this piece of code into its own function. It is now duplicated three times.

chaoran-chen (changes requested):

package.json
```diff
@@ -126,7 +126,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-puppeteer": "^6.0.0",
     "postcss": "^8.3.0",
-    "prettier": "^2.2.1",
+    "prettier": "^2.5.1",
```
chaoran-chen: Why are you changing the version of prettier?

src/pages/ExplorePage.tsx
gautier-collab: Right, it was a mistake.

src/pages/FocusPage.tsx
gautier-collab: Understood, I will do that.

package.json
gautier-collab: I had installed it manually, I just changed the prettier version to the original one.

chaoran-chen (approved):
Perfect! Feel free to merge :)

PR #415, opened 2022-02-28
Zero at plot bottom

### LINKED ISSUE #387: ENH: When line is 0% all the time, put ylim=0 at bottom rather than middle of plot
When a plot line is at 0% for the entire time, right now  ymax=-ymin, centering 0% in the middle of the plot, which is potentially confusing/misleading.

It may be better to put the 0% line somewhat at the bottom of the plot to show it's low rather than flat in the middle which could be misunderstood as ~50% or in any case not 0.

For illustration see this plot:
image

gautier-collab:
Issue  #387
Moved ylim=0 at the bottom rather than in the middle of plot when the curve is at 0 all the time. 
I did that for 4 charts.


chaoran-chen (approved):
Looks good to me!

PR #427, opened 2022-03-10
Feature/download fasta

### LINKED ISSUE #336: Download fasta
The user should be able to download the sequences (both aligned and unaligned) with the click of a button if the used data source is open.

It may make sense to wait for https://github.com/cevo-public/LAPIS/issues/5.

chaoran-chen:
This closes #336.

If the open version is used, it shows a dropdown menu below the "Sequence list (download)" button:

image

They redirect to e.g. https://lapis.cov-spectrum.org/open/v1/sample/fasta?host=Homo+sapiens&orderBy=random&limit=10000&region=Europe&dateFrom=2021-09-06&pangoLineage=BA.2*&downloadAsFile=true. As can be seen from the link, it sets a limit of 10,000. (LAPIS would allow up to 100,000).


corneliusroemer:
Wow! Thanks :)

PR #435, opened 2022-03-20
Show selected date ranges on subtitles

spacephoton:
CleanShot 2022-03-20 at 21 20 18
CleanShot 2022-03-20 at 21 22 03

### DIFF
--- a/src/widgets/VariantInternationalComparisonChart.tsx
+++ b/src/widgets/VariantInternationalComparisonChart.tsx
@@ -9,8 +9,6 @@ import { Utils } from '../services/Utils';
 import { CountryDateCountSampleEntry } from '../data/sample/CountryDateCountSampleEntry';
 import { fillFromWeeklyMap } from '../helpers/fill-missing';
 import { UnifiedIsoWeek } from '../helpers/date-cache';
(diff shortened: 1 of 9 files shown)


spacephoton:
do you know what the Jest issue could be ?

chaoran-chen:
I think that `react-bootstrap/esm/Form` is wrong. It should be `react-bootstrap/Form`.

src/widgets/VariantInternationalComparisonChart.tsx
```diff
@@ -109,6 +111,8 @@ export const VariantInternationalComparisonChart = ({
       isFixed: false,
     }))
   );
+  const exploreUrl = useExploreUrl();
```
chaoran-chen: The `...Chart` components should be (more or less) pure and are not supposed to fetch information through `useExploreUrl()`. Instead, they should only use information that is passed through the props. This is causing the current errors.

spacephoton:
Ah I see, so only this chart component right? was little confused so thanks!

anastasia-escher:
What's about this PR?
https://github.com/cevo-public/cov-spectrum-website/pull/435
It's been there for so long.

PR #440, opened 2022-03-25
Show rarer mutations

### LINKED ISSUE #433: ENH: Show mutations rarer than 5%, too
I think I've suggested this before but couldn't find the issue, maybe it was just a request for LAPIS but not for the website:

It would be really nice, if you showed the most common rare mutations for a given query too. Right now, only mutations with >5% share are shown. That means one misses out on interesting sub-lineages that may be growing but don't yet have a share of 5%.

One problem may be that this causes overwhelming numbers of mutations to show.

Possible solution:
Allow users to specify min (and maybe also max) proportion for mutations to show.

Looking for interesting mutations with say 1% share, I would then set min to 0.5%, max to 5% for example.

Maybe you could also add checkboxes to restrict to certain genes, e.g. S mutations only.

That would make the mutation widget on the website even more useful.

Just to be clear, I'm talking about this thing here:
image

I think given that LAPIS already supports min-frequency (at least for aa-mutations) this may be be a pure front-end task:
https://github.com/cevo-public/LAPIS/issues/9

dameyerdave:
Including this request would enable the user to chose the minProportion of mutations to be shown.

### DIFF
--- a/src/components/VariantMutations.tsx
+++ b/src/components/VariantMutations.tsx
@@ -12,6 +12,8 @@ import { SequenceType } from '../data/SequenceType';
 import { VariantSelector } from '../data/VariantSelector';
 import { PromiseQueue } from '../helpers/PromiseQueue';
 import { ReferenceGenomeService } from '../services/ReferenceGenomeService';
+import FormControl from 'react-bootstrap/FormControl';
+import InputGroup from 'react-bootstrap/InputGroup';
 
 export interface Props {
   selector: LocationDateVariantSelector;
@@ -63,6 +65,10 @@ export const VariantMutations = ({ selector }: Props) => {
         MutationProportionData.fromApi(selector, 'aa', signal),
         MutationProportionData.fromApi(selector, 'nuc', signal),
       ]).then(async ([variantCount, aaMutationDataset, nucMutationDataset]) => {
+
+        //console.log(aaMutationDataset)
+        console.log(nucMutationDataset)
+
         const aa = aaMutationDataset.payload;
         const nuc = nucMutationDataset.payload;
         const aaMap = new Map<string, MergedAAAndNucEntry[]>();
@@ -137,6 +143,10 @@ export const VariantMutations = ({ selector }: Props) => {
   }
   const data = queryStatus.data;
 
+  //console.log(data.aa)
+  //console.log( data.aa.filter((aa: any) => aa.proportion < 0.05 ) )
+  
+
   return (
     <>
       <div className='ml-4 mb-4'>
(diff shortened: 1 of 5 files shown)


src/data/api-lapis.ts
```diff
@@ -116,7 +116,15 @@ export async function fetchMutationProportions(
...
 ): Promise<MutationProportionEntry[]> {
-  const url = await getLinkTo(`${sequenceType}-mutations`, selector, undefined, undefined, undefined, true);
+  const url = await getLinkTo(
+    `${sequenceType}-mutations`,
+    selector,
+    undefined,
+    undefined,
+    undefined,
+    true,
+    '0.01'
```
chaoran-chen: The minimal selectable proportion is 0.001.

src/components/VariantMutations.tsx
```diff
@@ -122,12 +125,13 @@ export const VariantMutations = ({ selector }: Props) => {
     const nucFetchQueue = new PromiseQueue();
     for (let nucElement of nuc) {
       nucFetchQueue.addTask(() =>
-        fetchUniquenessScore(nucElement, selector, variantCount, 'nuc').then(uniqueness =>
+        fetchUniquenessScore(nucElement, selector, variantCount, 'nuc').then(uniqueness => {
+          //console.log('flag 5')
```
chaoran-chen: Please remove the commented-out code (there are a few more instances below).

chaoran-chen (commented):
Hi @dameyerdave, thanks for the PR! It looks great! Here are two minor comments.

corneliusroemer:
This would resolve #433

Amazing thanks!

Would it be possible to make it a text field rather than a slider? Slider is not very scientific and fiddly to get right - and with a linear scale like right now, it's annoying to get differences like 0.02 and 0.03

One could make the slider scale logarithmic, but even then, why fiddle if you can type...

image

chaoran-chen:
I think that the initial design by Gautier with the two text fields is actually quite nice. @dameyerdave, would it be a lot of trouble to change to that?

dameyerdave:
This is no trouble I can change it and also support the upper limit.

src/data/api-lapis.ts
dameyerdave: I changed this to `0.001`

src/components/VariantMutations.tsx
dameyerdave: I did so

chaoran-chen (approved):
Thanks @dameyerdave, this looks good! I just removed two other instances of commented-out code.

---

The view of the checkboxes is not so great on mobile, but this can be fixed another time :)

image

corneliusroemer:
Great, this is a big improvement and cool feature!

Typing in the field is maybe a bit awkward but I'll get used to it :)

PR #458, opened 2022-04-13
add dropdown for selecting geo

### LINKED ISSUE #321: Add resolution selection for "Geographic distribution" box
As it is currently, the geographic distribution box is suited for analysis of a particular country, but if the filter is set for `region` or `global` it becomes overwhelming. It would be nice to default to only showing the geographic division below the filter (e.g. `global->region`, `region->country`, `country->division`, `division->location`) but even better if there could be a selector (as the increased resolution may be useful for export).

Here is an example, filtered for "Europe":
image

anastasia-escher:
@chaoran-chen please check if it is what you meant :)


chaoran-chen:
Thanks, @anastasia-escher, this looks good! Just two small ideas for improvement:

- It would be nice if the tables could be sorted the same way as you did for the international comparison table
- Instead of showing "Granularity", I think that it would look better if it shows the granularity that is currently selected.

anastasia-escher:
Ok!

chaoran-chen (approved):
Thanks @anastasia-escher!

src/data/sample/DivisionCountSampleDataset.ts
```diff
@@ -14,25 +14,37 @@ export class DivisionCountSampleData {
     };
   }
 
-  static countByDivisionGroup(data: DivisionCountSampleEntry[]): Map<string, number> {
+  static countByDivisionGroup(
```
chaoran-chen, line 17: I think that the function should be either renamed or split up because `geoInfoArray` does not contain counts.

src/data/sample/DivisionCountSampleDataset.ts
```diff
@@ -44,5 +56,10 @@ export class DivisionCountSampleData {
         ];
       })
     );
+
+    return {
+      divisionData: divisionData,
+      geoInfoArray: DivisionCountSampleData.countByDivisionGroup(variant).geoInfoArray,
```
chaoran-chen, line 62: Why is `DivisionCountSampleData.countByDivisionGroup(variant)` being called again? It is already called in line 44.

src/helpers/geoAgg.ts
```diff
@@ -0,0 +1,61 @@
+export function aggregateGeo(
+  option: string,
```
chaoran-chen, line 2: `option` should be better typed as `'Divisions' | 'Countries' | 'Regions'`.

src/helpers/geoAgg.ts
```diff
@@ -0,0 +1,61 @@
...
+): Map<string, { count: number; prevalence: number }> {
+  const calc = new Map<string, { whole: number; variant: number }>();
+  const output = new Map<string, { count: number; prevalence: number }>();
+
+  switch (option) {
+    case 'Regions':
+      for (const item of data) {
+        if (item.country && item.region) {
+          if (!calc.has(item.region)) {
+            let whole: number = item.prevalence ? item.count / item.prevalence : 0;
```
chaoran-chen, line 19: `item.count / item.prevalence` looks inaccurate. It would be better to include the whole count in the input data. I am not sure if it could cause major errors in some cases (e.g., when the numbers are too small that they get approximated with 0?).

src/helpers/geoAgg.ts
```diff
@@ -0,0 +1,61 @@
...
+            calc.set(item.region, { whole: whole, variant: item.count });
+          } else {
+            const oldWhole = calc.get(item.region)?.whole ?? 0;
```
chaoran-chen, line 22: Another small thing: because you validated that `calc` contains `item.region` in the if clause above, I think that it is better to write `calc.get(item.region)!` instead of using `?`.

src/data/sample/DivisionCountSampleDataset.ts
```diff
@@ -14,25 +14,37 @@ export class DivisionCountSampleData {
...
+    data: DivisionCountSampleEntry[]
+  ): {
+    divisionData: Map<string, number>;
+    geoInfoArray: { country: string | null; region: string | null }[];
+  } {
+    let geoInfoArray: { country: string | null; region: string | null }[] = [];
     const output = new Map<string, number>();
     for (const entry of data) {
       if (entry.division === null) {
         continue;
```
chaoran-chen, line 27: **Source of error!** This falsifies the country and region counts if a country/region does not associate every sequence with a division (which they don't do). If a country (for a given variant) does not provide any division information, that country will not show up in the table at all!

src/widgets/VariantDivisionDistributionChart.tsx
```diff
@@ -32,70 +35,216 @@ const countriesWithMaps = [
...
+        : entries.sort((a, b) => (b.division > a.division ? 1 : a.division > b.division ? -1 : 0));
+  }
+};
+
 export const VariantDivisionDistributionChart = ({
   variantSampleSet,
   wholeSampleSet,
 }: VariantDivisionDistributionChartProps) => {
+  const [geoOption, setGeoOption] = useState<string>('Countries');
+  const [sortOption, setSortOption] = useState<SortOptions>('division');
```
chaoran-chen, line 79: It is strange to use lower case as internal value for one variable and upper case for another. Please try to be consistent (and use lower case).

src/widgets/VariantDivisionDistributionChart.tsx
```diff
@@ -32,70 +35,216 @@ const countriesWithMaps = [
...
+    const geoData = DivisionCountSampleData.proportionByDivision(
+      variantSampleSet.payload,
+      wholeSampleSet.payload
+    ).geoInfoArray;
+
     const variantDivisions = DivisionCountSampleData.proportionByDivision(
       variantSampleSet.payload,
       wholeSampleSet.payload
-    );
+    ).divisionData;
```
chaoran-chen, line 96: Why are you calling the same function with the same arguments twice?

src/widgets/VariantDivisionDistributionChart.tsx
```diff
@@ -32,70 +35,216 @@ const countriesWithMaps = [
...
+    case 'prevalence':
+      return ifAscendic
+        ? entries.sort((a, b) => a.prevalence - b.prevalence)
+        : entries.sort((a, b) => b.prevalence - a.prevalence);
+    case 'division':
+      return ifAscendic
+        ? entries.sort((a, b) => (a.division > b.division ? 1 : b.division > a.division ? -1 : 0))
+        : entries.sort((a, b) => (b.division > a.division ? 1 : a.division > b.division ? -1 : 0));
+  }
+};
```
chaoran-chen, line 72: Please use a `sortBy` helper function

src/widgets/VariantDivisionDistributionChart.tsx
```diff
@@ -32,70 +35,216 @@ const countriesWithMaps = [
...
+        setSortOption('count');
+        setIfAscCount(!ifAscCount);
+        setIfAscending(ifAscCount);
+        sortedAggData = sortGeoTable(sortedAggData, 'count', ifAscending);
+        break;
+      case 'prevalence':
+        setSortOption('prevalence');
+        setIfAscPrevalence(!ifAscPrevalence);
+        setIfAscending(ifAscPrevalence);
+        sortedAggData = sortGeoTable(sortedAggData, 'prevalence', ifAscending);
```
chaoran-chen, line 170: That's essentially redundant code.

src/widgets/VariantDivisionDistributionChart.tsx
```diff
@@ -32,70 +35,216 @@ const countriesWithMaps = [
...
     <DownloadWrapper name='VariantDivisionDistributionChart' csvData={csvData}>
+      {variantSampleSet.selector.location.country && country && !countriesWithMaps.includes(country) ? (
+        <DropdownButton id='dropdown-basic-button' title={geoOption}>
+          {granualityCountryOptions}
+        </DropdownButton>
+      ) : !country ? (
+        <DropdownButton id='dropdown-basic-button' title={geoOption}>
+          {granualityWorldOptions}
+        </DropdownButton>
+      ) : null}
```
chaoran-chen, line 209: This should also be written more concisely..

PR #465, opened 2022-04-19
Country selection

### LINKED ISSUE #459: Improve country selection
From the discussion in #321, I agree with @anastasia-escher that the current country selection in the top bar looks a bit unstructured. Maybe we can do something like this? We can then also add flags next to the country names.

image

anastasia-escher:
Reformatting the country select field with the material-ui library.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,11 @@
       "version": "0.1.0",
       "dependencies": {
         "@craco/craco": "^6.1.2",
+        "@emotion/react": "^11.9.0",
+        "@emotion/styled": "^11.8.1",
         "@fast-csv/format": "^4.3.5",
+        "@material-ui/core": "^4.12.4",
+        "@mui/material": "^5.6.2",
         "@sentry/react": "^6.16.1",
         "@sentry/tracing": "^6.16.1",
         "@testing-library/jest-dom": "^5.11.9",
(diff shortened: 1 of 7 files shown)


chaoran-chen:
Hi @anastasia-escher! Thanks for the very quick work!

I will look into the code in more detail tomorrow or on Thursday, but already a few quick comments:

- You mixed the geographic granularity code from #458 into this PR. It would however be better to keep the branches separated and base off from the main branch for every PR. This would make the reviews much easier because right now, it's difficult for me to distinguish the code for the two features. Could you please resubmit the PR with only the code for the country selection (do you know how to do it with git?).
- It seems a bit too much to include a big UI library just for a small feature. If you could find a way to implement this without adding a (or four) new dependencies, it would be perfect. But if not, since we had been quite generous with adding new libraries so far, I guess that we can also accept a few more...
- Is this intended? It looks a bit weird to me that the field border does not go around the whole field, but I assume that this can also be interpreted as design.
image

chaoran-chen:
Visually, it is now looking great! But the actual functionality (the country switching) is not working for me anymore. Is it working for you, @anastasia-escher?

anastasia-escher:
It works differently, you have to change the country and click somewhere outside of the field (I had to use onBlur instead of onChange so that the value doesn't change with every typed letter but after the typing is done).

chaoran-chen:
Ah! Could you maybe also implement switching by clicking the enter button?

anastasia-escher:
Now it should switch directly by selecting.

chaoran-chen:
Perfect!

I have two more ideas:

- Remove/reduce the left margin (red)
- Align the country names even if there is no flag available (blue)

image

chaoran-chen (approved):
Looks good to me! Thank you! Please merge if you think it's ready :)

PR #473, opened 2022-04-25
fix linking issue

### LINKED ISSUE #471: Links don't keep country/continent selection
From the last update i am noticing that links to a query in a determinated country or area don't keep the selection for a choosen country and instead they link to default "switzerland" page.

it could be a big issue for pango proposals forcing the pango team to reconstruct what that link means.
it is the same for comparison btw variants pages. and it is true for old links too 

So this could influence a lot of things.

(for example i have a rapid list to check growth advantages or nr of sequencss, and it is not rapid at all now)

anastasia-escher:


chaoran-chen:
Hi. Sorry for the slow response. The linking is now working but the location field is empty by default (if you don't put something in):

image

chaoran-chen (approved):
Looks good, thank you!

PR #477, opened 2022-05-01
date picker issue

### LINKED ISSUE #441: BUG/ENH: Start date cannot be typed, text entry field misbehaves
When I try to edit the string corresponding to the date range selection, the text field misbehaves making it impossible to edit the start date directly by keyboard.

It would be nice if start and end date were in two separate fields so that they could both be edited by keyboard.

With the current date picker, it's quite tedious to quickly do something like Jun 2020 to Jan 2021, because one needs to click once for every month going backwards.

image

Splitting in two boxes may also help with #377

anastasia-escher:
Added separate fields for start and end dates.


chaoran-chen:
Thanks @anastasia-escher!

@corneliusroemer, please check this out! Does this work well for you or do you have further suggestions?

chaoran-chen (changes requested):
The mobile view is unfortunately now even worse than before. The dropdown menu for the preset date ranges (All times, past 3 months, past 6 months, etc.) is now too small.

image

corneliusroemer:
Thanks @anastasia-escher and @chaoran-chen, that's great, now start and end dates can be typed!

Together with #485 this should resolve any date selection UX issues I'm having

corneliusroemer:
Hmm, now that I tested it, found a bug: If you type a date in the field and press enter, the new typed date is not accepted. One needs to click in the date picker, which shouldn't be necessary (I think)

corneliusroemer:
I think it's fixed by [0be1227](https://github.com/cevo-public/cov-spectrum-website/pull/477/commits/0be122754149643f28f8109538e54713c24ab421) now, thanks @anastasia-escher

chaoran-chen:
Hi. Sorry for the late response. Something is still wrong on my device. When I change the date with the date picker, it does not have any effect. Only when I choose the date with the date picker a second time, it works.

anastasia-escher:
Hi @chaoran-chen! And if you pick a date for the first time and press enter?

anastasia-escher:
@corneliusroemer , @chaoran-chen  please review

corneliusroemer:
@anastasia-escher thanks! There's a pretty big bug if you don't type a `-` after say `2022` but add a number, say typing `20223` in the date field.

As a result, the website goes completely white. Can you reproduce that? If not, I'm happy to give more details.

The usual UI flow for dates is that `-` are added automatically, maybe there's a material UI date field like that that could be used?

anastasia-escher:
@corneliusroemer , @chaoran-chen this is now a MUI version of the date picker, please check it out.

package.json
```diff
@@ -4,11 +4,16 @@
   "private": true,
   "dependencies": {
     "@craco/craco": "^6.1.2",
+    "@date-io/date-fns": "^2.14.0",
+    "@date-io/dayjs": "^2.14.0",
+    "@date-io/luxon": "^2.14.0",
+    "@date-io/moment": "^2.14.0",
```
chaoran-chen: Hi @anastasia-escher. I will check out everything else later but already have a question: why are the above four dependencies needed?

corneliusroemer:
Looks really good @anastasia-escher!
I think it'd be good to switch the date format to ISO date like `2022-01-22` but that shouldn't be too hard I hope.

Really nice and reactive now, no need to press enter, easy to navigate with keyboard.

chaoran-chen:
@anastasia-escher, this looks good to me as well! Just like @corneliusroemer, I think that it would be better to use dates in the ISO format.

src/components/DateRangePicker.tsx
```diff
@@ -78,6 +78,7 @@ export const DateRangePicker = ({ dateRangeSelector }: Props) => {
         <div className={`flex flex-row ${width && width < 600 ? 'flex-wrap mt-2 mb-2 ml-0' : 'ml-2'}`}>
           <LocalizationProvider dateAdapter={AdapterDateFns}>
             <DatePicker
+              inputFormat='MM-dd-yyyy'
```
chaoran-chen: @anastasia-escher, ISO 8601 (https://www.iso.org/iso-8601-date-and-time-format.html) defines the date format as `yyyy-MM-dd` ;)

chaoran-chen:
Another thing: We used the package `react-datepicker` before. Can it now be removed?

PR #481, opened 2022-05-02
add GISAID list

### LINKED ISSUE #444: ENH: Provide simple GISAID accession list
Right now, information about sequences returned by a query can be downloaded.

But the format is a CSV with multiple columns.

To download sequences from GISAID, it's thus necessary to process the CSV manually to get just a list of accession IDs.

It would be nice if one could get a simple new line separated list of EPI_ISLs at the click of a button.

Ideally as plaintext in a new tab, so that one can copy directly, without having to open in an editor.

Random order would be good, too, so that one gets representative excerpt without randomisation.

anastasia-escher:
add a button-link to a plain text page with the GISAID list

### DIFF
--- a/src/components/FocusVariantHeaderControls.tsx
+++ b/src/components/FocusVariantHeaderControls.tsx
@@ -81,9 +81,11 @@ export const FocusVariantHeaderControls = React.memo(
 
           <Dropdown.Toggle split variant='secondary' />
           <Dropdown.Menu>
-            <ExternalLink url={listLink2 ?? ''}>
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Splendid! Thanks @anastasia-escher, this is a really really useful feature that many people are going to appreciate.

I tested and it works great! So little code and such great benefit :) @chaoran-chen 

closes #444

chaoran-chen (approved):
Thanks, @anastasia-escher!

PR #482, opened 2022-05-02
add toggle for showing deletions

### LINKED ISSUE #466: ENH: Checkbox to toggle deletion visibility in mutation view
Due to artefacts and the Nextclade trailing deletion issue, when trying to find mutations arising, that are present at between 0.1-5%, the overwhelming number if mutations displayed are deletions.

Now deletions are actually the only thing I'm usually not interested in, since they are much less informative (due to artifacts and lack of uniqueness even if genuine).

So it would be great if it was possible to hide all deletions - through a checkbox. That would improve usability A LOT.

Maybe it would even make sense to hide deletions by default and show only upon toggle. Or if it was remembered for queries through a cookie so that one doesn't need to reselect for every single query.

I've used a workaround till now of including `ORF1a:. , N:420.` but that's really suboptimal to this solution. As it filters out sequences and makes queries more lengthy than necessary. And doesn't solve problem of long interior deletion stretches.

anastasia-escher:
add toggle for showing deletions

### DIFF
--- a/src/components/VariantMutations.tsx
+++ b/src/components/VariantMutations.tsx
@@ -14,6 +14,9 @@ import { ReferenceGenomeService } from '../services/ReferenceGenomeService';
 import NumericInput from 'react-numeric-input';
 import { LapisSelector } from '../data/LapisSelector';
 import { useResizeDetector } from 'react-resize-detector';
+import Checkbox from '@mui/material/Checkbox';
+import FormGroup from '@mui/material/FormGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
 
 export interface Props {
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Amazing, so useful! Looking forward for this to be merged.

Tested it quickly and seems to work as expected.

Btw, @anastasia-escher if you write `resolves #466` somewhere in the pull request message, GISAID will magically assign this PR to the issue and close if merged :)

anastasia-escher:
Thanks for the hint, @corneliusroemer, I didn't know that!

chaoran-chen (approved):
Looks great, thanks!

PR #485, opened 2022-05-02
feat: add past month, and past 2 months as preset date ranges

corneliusroemer:
Partially resolves #442

Parallel changes to 209245ec208e7375daf6eea24caeec7ae1c9497d (add date selector for 2022)

Let's see whether this works in Vercel 🤞

PR #488, opened 2022-05-04
feat: add 2 weeks date selector preset

corneliusroemer:
Adds recent 2 week date selector, parallel to #485


corneliusroemer:
Tested and works

chaoran-chen (approved):
Okay :) But when would you need that? The number of samples from the past two weeks are usually quite low, I assume?

corneliusroemer:
They are low, but they are the most sensitive/informative for new developments/variants.

I use it a lot to look at very recent developments in mutation view

PR #491, opened 2022-05-06
478 venn remove code redundancy

### LINKED ISSUE #478: Reduce code redundancy in SvgVennDiagrams
There are still a lot of code redundancies in the three `SvgVennDiagram`s that should be removed.

anastasia-escher:


chaoran-chen:
Hi @anastasia-escher, thanks for implementing this! I didn't anticipate it but now I noticed that for intersections with many mutations, the tooltip is huge and pretty much shadows everything. Then, the colorings are not very useful, but I'm not sure how to improve this. Do you have some ideas?

image

anastasia-escher:
Hi @chaoran-chen, the tooltip could appear at click, would it be ok?

chaoran-chen:
I think that seeing the mutations is very relevant and having to click every time is too much.

anastasia-escher:
Then, maybe we can configure the background color of the tooltip to be more transparent?

anastasia-escher:
And to to make the venn4 diagram bigger.

chaoran-chen:
What do you think about the following idea:

**Remove tooltips:** When the user hovers over a field, instead of showing the mutations in a tooltip, they could be printed below the chart - like this:

**Highlight the variant labels:** Instead of highlighting many areas (which I find a bit confusing), what's about underlining the variant labels? And only the area, that is currently hovered on, will be colored - with the same color as the lines under the variant labels.

**Hover over area:** It should be enough to hover over an area to select it. Currently, the user has to hover over the texts.

---

Together, it could look like this:

image

anastasia-escher:
Yes, this sounds good!

chaoran-chen (approved):

chaoran-chen:
Hi @anastasia-escher. I made a few changes, could you please check if it is now working as expected? Here are my changes: https://github.com/cevo-public/cov-spectrum-website/pull/491/files/122e2749f9c37cdf8d54248b225755db45f667e4..9918a9a3255c08b6c527c75cb9590ed44a6f2b75

I was able to remove around 300 lines of redundant code. (And there is still quite some redundancy so we could probably remove another 50-100 lines.) Please merge if you think that it's fine!

PR #500, opened 2022-05-15
add a static legend to the map

### LINKED ISSUE #291: [Enhancement] Geographic plots should have a legend for color -> percentage
It would be great if geographic plots had a legend showing which number a color corresponds to:
image

Hovering by mouse isn't really that great for screenshots etc.

anastasia-escher:
Regarding #291, add a static legend to the map.


chaoran-chen (approved):
Nice, thank you! That fulfills its purpose. Please fix the tests and then, feel free to merge.

chaoran-chen:
Oh, and by the way, please write "Closes #291" or "Resolves #291" or "Fixes #291" - not e.g. Regarding because only those keywords would close the issue after merging a PR.

corneliusroemer:
Why do the tests here fail?

Looks good. We still would want to prevent the colour scale from being discretized, though. The legend is helpful though :) 

Would be great if you could add a screenshot to the PR whenever you open one so that it's clear at a glance what is being changed and what it looks like, a bit like this.

image

chaoran-chen:
I will look at the tests.

PR #501, opened 2022-05-16
Update Taxonium integration to Taxonium v2

theosanderson:
Sorry Taxonium has changed and now uses a different URL (although there is currently a redirect URL so things still work). I've now tested this locally and it seems to work. As you've mentioned previously it would be much better with AND/OR searches. The new architecture approach brings that closer, but it's not here yet.


chaoran-chen (approved):
Thank you very much, Theo!

PR #502, opened 2022-05-18
add alternative ways to specify ORF1a or ORF1b

### LINKED ISSUE #496: Auto-completion: Translate NSP and ORF1ab notation to ORF1a/b
There are two alternative ways to specify ORF1a and ORF1b AA mutations: either by seeing ORF1ab as one single gene or distinguish between the different nsps. E.g., `ORF1b:12` is equivalent to `ORF1ab:4413` and `nsp12:21`. There is a nice tool by @theosanderson (and @zach-hensel) to do the conversion (https://codon2nucleotide.theo.io/) and its code reveals how the conversion can be done (https://github.com/theosanderson/Codon2Nucleotide/blob/main/src/App.js).

Internally (and from the perspective of the APIs), we will only support the `ORF1a` and `ORF1b` notation. But to make it easier for users, it would be nice if our search bar supports the other notations as well and automatically translates them into our notation. I.e., if the user enters either `ORF1ab:4413L` or `nsp12:21L`, the auto-completion should suggest `ORF1b:12L`.

image

Further, if the user copy&pastes a comma-separated string like `nsp12:21L,nsp12:22P` into the search bar, it should convert the input to `ORF1b:12L` and `ORF1b:13P`.

GISAID uses the nsp notation and this feature would make it much easier to copy the mutations of a sequence shown on GISAID into cov-spectrum.

anastasia-escher:
resolves #496 (probably)

### DIFF
--- a/src/components/VariantSearchField.tsx
+++ b/src/components/VariantSearchField.tsx
@@ -26,9 +26,14 @@ const backgroundColor: { [key in SearchType]: string } = {
 };
 
 function mapOption(optionString: string, type: SearchType): SearchOption {
+  let actualValue = optionString;
+  if (optionString.includes('(=')) {
+    actualValue = optionString.split(' ')[0];
+  }
+
   return {
     label: optionString,
-    value: optionString,
+    value: actualValue,
     type,
   };
 }
@@ -159,6 +164,7 @@ export const VariantSearchField = ({ onVariantSelect, currentSelection, triggerS
   const suggestOptions = (query: string): SearchOption[] => {
     const onePLAlreadySelected = selectedOptions.filter(option => option.type === 'pango-lineage').length > 0;
     const suggestions: SearchOption[] = [];
+
     if (isValidAAMutation(query)) {
       suggestions.push(mapOption(query, 'aa-mutation'));
     } else if (isValidNucMutation(query)) {
@@ -203,8 +209,10 @@ export const VariantSearchField = ({ onVariantSelect, currentSelection, triggerS
   }
 
   const translateMutation = (oldValue: string) => {
-    let nsp: string | null = null;
+    let orf1aorb: string | null = null;
     let orf1aorbcodon: number | null = null;
+    let nsp: string | null = null;
+
     let nspCodon: number | null = null;
     let combinedCodon: number | null = 3646;
     let mutationArray = oldValue.split(':');
@@ -214,7 +222,6 @@ export const VariantSearchField = ({ onVariantSelect, currentSelection, triggerS
       let letterAfter = mutationArray[1].charAt(mutationArray[1].length - 1);
       orf1aorbcodon = parseInt(mutationArray[1].slice(1, -1));
       combinedCodon = orf1aorbcodon + (mutationArray[0] === 'ORF1b' ? 4401 : 0);
-      console.log(combinedCodon, orf1aorbcodon);
       nsp = 'nsp1';
       nspCodon = combinedCodon - nsps[nsp] + 1;
       for (const [key, value] of Object.entries(nsps)) {
@@ -225,9 +232,35 @@ export const VariantSearchField = ({ onVariantSelect, currentSelection, triggerS
       }
 
       return `${nsp}:${letterBefore}${nspCodon}${letterAfter} = ORF1ab:${letterBefore}:${combinedCodon}${letterAfter}`;
+    } else if (mutationArray[0].toLowerCase().startsWith('nsp')) {
+      nspCodon = is_numeric(mutationArray[1].charAt(0))
+        ? parseInt(mutationArray[1].slice(0, -1))
+        : parseInt(mutationArray[1].slice(1, -1));
+      console.log(nspCodon);
+      nsp = mutationArray[0];
+      combinedCodon = nsps[nsp] + nspCodon - 1;
+      if (combinedCodon >= nsps['nsp13']) {
+        orf1aorbcodon = nsps[nsp] + nspCodon - 1 - 4401;
+        orf1aorb = 'ORF1b';
+      } else {
+        orf1aorbcodon = nsps[nsp] + nspCodon - 1;
+        orf1aorb = 'ORF1a';
+      }
+      return orf1aorb + ':' + orf1aorbcodon;
+    } else if (mutationArray[0].toLocaleLowerCase() === 'orf1ab') {
+      combinedCodon = parseInt(mutationArray[1].slice(0, -1));
+      if (combinedCodon > 4401) {
+        orf1aorbcodon = combinedCodon - 4401;
+        orf1aorb = 'ORF1b';
+      } else {
+        orf1aorb = 'ORF1a';
+        orf1aorbcodon = combinedCodon;
+      }
+      return orf1aorb + ':' + orf1aorbcodon;
     }
-    return '';
+    return oldValue;
   };
+
   // nsp + ':' + letterBefore + nspCodon; //
   /**
    * Handles the input change:
@@ -297,7 +330,6 @@ export const VariantSearchField = ({ onVariantSelect, currentSelection, triggerS
     // remove the last "," in the invalidQueries string
     invalidQueries = invalidQueries.slice(0, -1);
     setInputValue(invalidQueries);
-
     setMenuIsOpen(invalidQueries !== '');
   };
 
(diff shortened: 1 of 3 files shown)


chaoran-chen:
I don't think that it is already fully working. In the following example, it is not translating?

image

anastasia-escher:
Oh, it works now only with lower case - nsp, so I will make it case-independent and recommit .

anastasia-escher:
Hm, but now I see that it works also with upper case for me, strange. 
Screenshot_20220519-224904_DuckDuckGo

anastasia-escher:
Ah, now I understand, it only shows the alternative in the autocomolete options if you type, but it doesn't translate if you paste. I will change it.

anastasia-escher:
@chaoran-chen please check if it works correctly now.

chaoran-chen:
@anastasia-escher, it is now translating after pressing enter but it would also be nice if it translates the autosuggestion as well:
image

Further, if someone enters a NSP... value, it does not need to show the ORFab equivalent. And the same when someone enters the ORFab value. We only need a translation to ORFa/ORFb.

image

anastasia-escher:
@chaoran-chen please check if it works correctly now.

chaoran-chen:
Sorry if I was still unclear: When the user enters "NSP...", then it should show something like "ORF... (= NSP...)". When the user enters "ORF1ab ...", it should show "ORF... (= ORFab...)"

anastasia-escher:
@chaoran-chen and now?

chaoran-chen:
Still not getting an auto suggestion in every case? (First issue mentioned in https://github.com/cevo-public/cov-spectrum-website/pull/502#issuecomment-1134406011)

image

anastasia-escher:
@chaoran-chen now it should work

corneliusroemer:
Tested it a bit and it seems to work well :)

corneliusroemer:
@anastasia-escher you really need to start with the keywords: `resolves` `fixes` etc, if you add `probably` before Github won't link.

You don't need to be certain that it fixes it, it's enough that that's what you're trying to achieve.

So I edited your PR message to `resolves #496 (probably)` so now it's linked :)

chaoran-chen:
If I enter `NSP4:4K`, it gets translated to `ORF1a:3767`. The `K` got lost.

anastasia-escher:
> @anastasia-escher you really need to start with the keywords: `resolves` `fixes` etc, if you add `probably` before Github won't link.
> 
> You don't need to be certain that it fixes it, it's enough that that's what you're trying to achieve.
> 
> So I edited your PR message to `resolves #496 (probably)` so now it's linked :)

oh sorry, I always forget about it

chaoran-chen:
image


@anastasia-escher, there is no need to show the `ORF1ab`-equivalence if the user enters a mutation using our main notation (i.e., ORF1a or ORF1b). Just to clarify again, we only show translations in one direction: from the NSP notation and the ORF1ab notation to our notation, not back.

chaoran-chen:
Another small inconsistency/strange behavior: When entering a NSP value, the (optional) character before the position disappears:

image

This does not happen when entering an ORF1ab value:

image

src/components/VariantSearchField.tsx
```diff
@@ -25,10 +25,157 @@ const backgroundColor: { [key in SearchType]: string } = {
...
+  nsp9: 4141,
+  nsp10: 4254,
+  nsp12: 4393,
+  nsp13: 5325,
+  nsp14: 5926,
+  nsp15: 6453,
+  nsp16: 6799,
+};
+
+function getEquivalent(value: string, notation: string) {
```
chaoran-chen: - It would be better to narrow down the type of `notation`: I think that it could be `'nsp' | 'orf1ab'`?
- I think that this function's logic is not obvious and it would be great if you could document/explain the algorithm a bit.

src/components/VariantSearchField.tsx
```diff
@@ -25,10 +25,157 @@ const backgroundColor: { [key in SearchType]: string } = {
...
+  let orf1aorbcodon: number | null = null;
+  let nsp: string | null = null;
+  let nspCodon: number | null = null;
```
chaoran-chen: Is there a reason why you initiate the three variables above already here and not when they are needed?
In which case do they need to be `null`?

src/components/VariantSearchField.tsx
```diff
@@ -25,10 +25,157 @@ const backgroundColor: { [key in SearchType]: string } = {
...
+  let combinedCodon: number | null = 3646;
```
chaoran-chen: In which case could `combinedCodon` be `null`?

src/components/VariantSearchField.tsx
```diff
@@ -25,10 +25,157 @@ const backgroundColor: { [key in SearchType]: string } = {
...
+      return `${nsp}:${letterBefore}${nspCodon}${letterAfter}`;
+    } else if (notation === 'orf1ab') {
+      return `ORF1ab:${letterBefore}:${combinedCodon}${letterAfter}`;
+    } else {
+      return value;
+    }
+  }
+}
+
+const translateMutation = (oldValue: string) => {
```
chaoran-chen: This function shares some code with `getEquivalent()`. Is it possible to reduce code redundancy?

src/components/VariantSearchField.tsx
```diff
@@ -101,36 +248,37 @@ export const VariantSearchField = ({ onVariantSelect, currentSelection, triggerS
   const suggestMutations = (query: string): string[] => {
     // TODO Fetch all/common known mutations from the server
     // For now, just providing a few mutations so that the auto-complete list is not entirely empty.
-    return [
+    let notation = query.toLowerCase().startsWith('n') ? 'nsp' : query.startsWith('orf1ab') ? 'orf1ab' : '';
+    let options: string[] = [
       'S:D614G',
-      'ORF1b:P314L',
+      `ORF1b:P314L`,
```
chaoran-chen: To keep the PR clean, please try to avoid unnecessary edits and revert the changes from ' to `

src/components/VariantSearchField.tsx
```diff
@@ -25,10 +25,157 @@ const backgroundColor: { [key in SearchType]: string } = {
...
+const translateMutation = (oldValue: string) => {
+  let orf1aorb: string | null = null;
+  let orf1aorbcodon: number | null = null;
+  let nsp: string | null = null;
+
+  let nspCodon: number | null = null;
+  let combinedCodon: number | null = 3646;
+  let mutationArray = oldValue.split(':');
+
+  if (mutationArray[0] === 'ORF1b' || mutationArray[0] === 'ORF1a') {
```
chaoran-chen: Is this part of the function outdated? Since we don't need to translate ORF1a/ORF1b mutations into another notation?

src/components/VariantSearchField.tsx
```diff
@@ -25,10 +25,157 @@ const backgroundColor: { [key in SearchType]: string } = {
...
+        ? parseInt(mutationArray[1].slice(0, -1))
+        : parseInt(mutationArray[1].slice(1, -1));
+    } else {
+      nspCodon = is_numeric(mutationArray[1].charAt(0))
+        ? parseInt(mutationArray[1])
+        : parseInt(mutationArray[1].slice(1));
+    }
+
+    nsp = mutationArray[0];
+    combinedCodon = nsps[nsp] + nspCodon - 1;
```
chaoran-chen: The auto-suggestion should show nothing if the user types a non-existing NSP and should not translate it to `ORF1a:NaN`:

image

src/components/VariantSearchField.tsx
```diff
@@ -25,10 +25,157 @@ const backgroundColor: { [key in SearchType]: string } = {
...
+};
+
 function mapOption(optionString: string, type: SearchType): SearchOption {
+  let actualValue = optionString;
+
+  if (optionString.includes(' =')) {
+    actualValue = optionString.split(' ')[0];
+  } else if (optionString.toLowerCase().startsWith('nsp')) {
+    if (translateMutation(optionString.toLocaleLowerCase())) {
+      actualValue = translateMutation(optionString.toLocaleLowerCase());
```
chaoran-chen: Is it intended to use `toLocaleLowerCase()` instead of `toLowerCase()` like everywhere else here?

chaoran-chen (changes requested):
Hi @anastasia-escher, I'm sorry for the slow response times! I left a number of comments in the code. In addition, the parsing of comma and plus-sign separated lists of values does not work yet.

If I paste `nsp6:77,ORF1ab:A500G,ORF1b:34,nsp14:320T` into the search field, only ORFab:34 gets parsed but the others are not getting translated.

chaoran-chen (approved):
Thank you @anastasia-escher, this looks great!

Issue #462, opened 2022-04-14
ENH: Optional logit scale for share plots


corneliusroemer:
It would be great if covSPECTRUM had an optional logit scale button for share plots.

It would make comparisons like these useful. Right now there's not much to be seen.

With logit axis, it'd be super useful.

Slope as growth advantage ;)

I think this should be doable and would be a huge benefit!

![Screenshot_20220414-150719.jpg](https://user-images.githubusercontent.com/25161793/163397470-5ed52d87-16e6-4358-8f28-80a7f23fab7f.jpg)

KarelBerka:
that would be a great improvement as it is useful for detecting rising slopes even in small numbers (as was visible in absolute numbers for Czech cases over all summer, while the total numbers kept the public completely unaware of the autumn Delta spree).

chaoran-chen:
I agree, it would be useful to have it for the "Sequences over time" plots both in the single and the two comparison modes.

@anastasia-escher, may I assign this to you as one of the upcoming tasks? :)

anastasia-escher:
@chaoran-chen sure!

anastasia-escher:
Hi @chaoran-chen and @corneliusroemer, please check this
https://cov-spectrum-develop-3wb4lcdlb-cov-spectrum.vercel.app/explore/Switzerland/AllSamples/Past6M/variants?variantQuery=BA.3*&variantQuery1=BA.1*&analysisMode=CompareEquals&

chaoran-chen:
<img width="941" alt="image" src="https://user-images.githubusercontent.com/18666552/180654230-02b1d3ae-da3d-478f-b135-5ac5038d79e8.png">

This doesn't look right, why are there negative values? The y-axis should still be in the range of 0 to 100%. (log(0) should be considered as undefined and not shown.) The implementation should be equivalent to the "log scale" mode for the international comparison. See: https://cov-spectrum-develop-3wb4lcdlb-cov-spectrum.vercel.app/explore/Switzerland/AllSamples/Past6M/variants/international-comparison?variantQuery=BA.3*&variantQuery1=BA.1*&

chaoran-chen:
Also to clarify here: This issue is requesting a logit scale, not a log scale but I think that it would make sense to add both.

chaoran-chen:
By the way, a related issue is #211 which is asking for a log scale option in the estimated cases plot.

chaoran-chen:
[...]

PR #534, opened 2022-07-26
add log scale

anastasia-escher:
solves #462 

only log scale for now


package.json
```diff
@@ -76,7 +76,7 @@
   "scripts": {
     "start": "craco start",
     "build": "craco build",
-    "test": "craco test",
+    "test": "craco test -u",
```
chaoran-chen: Please don't change the default behavior of the test command.

chaoran-chen (changes requested):
Hi @anastasia-escher, thanks for the PR! The absolute plot is not correct. When you check the log scale option and switch to the "absolute" numbers, the log scale checkbox disappears but it seems that it is still on and the y-axis is wrong:

image

chaoran-chen (changes requested):
If I switch to the "Absolute" view and then switch back to "Proportion", the plot looks like this:

image

(Observed in both Chrome and Firefox on Windows)

chaoran-chen:
Hmm.. this also isn't ideal... there is now a lot of empty space:

image

Why is the problem occurring?

anastasia-escher:
> Hmm.. this also isn't ideal... there is now a lot of empty space:
> 
> image
> 
> Why is the problem occurring?

This has to do with the height of the plot wrapper and the margin-bottom of the plot... I'll try to make it look nicer.

chaoran-chen (changes requested):
In the compare mode, the plot is now very small. The Log scale checkbox should be placed in the same row as the show CI button if there is sufficient space. There also shouldn't be so much empty space at to bottom.

image

corneliusroemer:
This is nice already, though as @chaoran-chen mentioned, the button could be placed in the same row as `absolute` etc
image

chaoran-chen (changes requested):
The UI is unfortunately still not working properly.

On the mobile view, it's plot is still tiny:

image

It introduced a lot of space under many plots. This is how the growth advantage plot looks on the production page:

image


This is how this PR looks like:

image

chaoran-chen:
@dameyerdave, would you have the time to take a look and help out here?

src/widgets/common.tsx
```diff
@@ -40,6 +40,7 @@ export const ChartAndMetricsWrapper = styled.div`
   display: flex;
   flex-direction: column;
   flex: 1;
+  margin-bottom: 70px;
```
chaoran-chen: This line looks problematic. The `ChartAndMetricsWrapper` is a widely used component. Is it a good idea to set such a fixed margin here?

dameyerdave:
Hi Chaoran, I totally missed you question. I'm very sorry for that. @anastasia-escher is on this and will solve this.

chaoran-chen (approved):
Thanks!

PR #539, opened 2022-07-29
Lodash

### LINKED ISSUE #443: Remove lodash
As a small code improvement measure: the [lodash library](https://github.com/lodash/lodash) does not seem to be actively maintained anymore and as we only use it [a very few times](https://github.com/cevo-public/cov-spectrum-website/search?q=lodash), I think that we should get rid of it. The [you-dont-need repository](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_capitalize) might be helpful.

anastasia-escher:
resolves #443


src/components/CombinedExport/ExportButton.tsx
```diff
@@ -1,9 +1,13 @@
-import { sortBy } from 'lodash';
+//import { sortBy } from '../../helpers/lodash_alternatives';
```
chaoran-chen: I suppose that you would like to you the `sortBy` function in `lodash_alternatives` which is exactly the same as the in lines 7-9?

src/helpers/deep-compare-hooks.ts
```diff
@@ -1,10 +1,12 @@
 import { DependencyList, EffectCallback, useEffect, useRef, useState } from 'react';
-import _ from 'lodash';
+import isEqual from './isEqual';
 
 // Thanks https://stackoverflow.com/a/54096391/8376759
 function useDeepCompareMemoize<T>(value: T): T | undefined {
   const ref = useRef<T>();
-  if (!_.isEqual(value, ref.current)) {
+  console.log('value', value);
+  console.log('ref.current', ref.current);
```
chaoran-chen: Please remove console.logs

src/helpers/isEqual.js
```diff
@@ -0,0 +1,33 @@
+const isEqual = (first, second) => {
+  /* Checking if the types and values of the two arguments are the same. */
```
chaoran-chen: It's not very clear what "the same" means. I'd rather reword the comment and use the technical terminology and write "strictly equal". The strict equality is defined in the MDN docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality#description

src/helpers/isEqual.js
```diff
@@ -0,0 +1,33 @@
+const isEqual = (first, second) => {
```
chaoran-chen: - Why are you writing this in pure JavaScript (file ending is `.js`) and not in TypeScript like everything else?
- Please document the cases in which the function works/is supposed to be used. E.g., I believe that it will not work properly for `ArrayBuffer`s?

src/helpers/isEqual.js
```diff
@@ -0,0 +1,33 @@
...
+  /* Checking if the types and values of the two arguments are the same. */
+  if (first === second) return true;
+
+  /* Checking if any arguments are null */
+  if (first === null || second === null) return false;
```
chaoran-chen: What's about `undefined`? Why does that not get a similar special handling?

src/helpers/isEqual.js
```diff
@@ -0,0 +1,33 @@
...
+
+  /* Checking if any argument is none object */
+  if (typeof first !== 'object' || typeof second !== 'object') return false;
+
+  /* Using Object.getOwnPropertyNames() method to return the list of the objects’ properties */
+  let first_keys = Object.getOwnPropertyNames(first);
+
```
chaoran-chen: Remove the new line here?

src/helpers/lodash_alternatives.ts
```diff
@@ -0,0 +1,54 @@
+import { UnifiedIsoWeek } from './date-cache';
+
```
chaoran-chen: I think that all the functions in this file deserve some documentation. If a function works exactly as in `lodash`, it might make sense to add a link to the lodash documentation of the corresponding function. If not, please document how it exactly works and maybe also point out the difference to lodash.

src/pages/FocusSinglePage.tsx
```diff
@@ -59,13 +59,30 @@ export const FocusSinglePage = () => {
...
+    }, {});
+
+    console.log(res);
+
+    return res;
+  }
+
+  const deepFocusButtons = useMemo(() => mapValues(deepFocusPaths), [exploreUrl]);
+
+  console.log('deepFocusButtons', deepFocusButtons);
```
chaoran-chen: What happened here? Why so many changes?

src/pages/FocusSinglePage.tsx
```diff
@@ -395,6 +412,7 @@ export const FocusSinglePage = () => {
                 <NamedCard
                   title='Relative growth advantage'
                   toolbar={[
+                    // @ts-ignore
```
chaoran-chen, line 399: Why that?

chaoran-chen (changes requested):
Hi @anastasia-escher and @dameyerdave, thanks for your work! It seems to work well and I couldn't find any error in the code.

In addition to the comments and questions in the code and as discussed in our last meeting, I think that we should write tests for all the helper functions here. As it is not urgent to merge this PR, I'd suggest adding the tests as part of this PR.

src/helpers/isEqual.js
anastasia-escher: > What's about `undefined`? Why does that not get a similar special handling?

Actually I am thinking about removing  this line. Both objects could very well be null or undefined and if the are, they are equal (i. e. we don't need to return false if both are null or undefined, we need to return true).  Does that make sense?

PR #554, opened 2022-08-10
align fields on focus page; rounded special date range button

anastasia-escher:
align fields on the focus page; rounded the special date range button


chaoran-chen (approved):
Looks great, thanks!

PR #556, opened 2022-08-10
Baseline comparison style

anastasia-escher:
Moved variant search field to the next line and adjusted the alignment.


chaoran-chen (commented):
I think that's a good idea and would like to ask for just a few small changes:

- I think that it would be better if the search in the "Compare variants" (without baseline) is also moved to a new line.
- The margin should be smaller between the search and the content. 
image

PR #559, opened 2022-08-11
CompareToBaseline and CompareEquals variant search field style

anastasia-escher:
Change styles for CompareToBaseline and CompareEquals variant search fields.

### DIFF
--- a/src/components/DateRangePicker.tsx
+++ b/src/components/DateRangePicker.tsx
@@ -85,11 +85,17 @@ export const DateRangePicker = ({ dateRangeSelector }: Props) => {
                 }
                 setStartDate(value);
               }}
-              renderInput={params => <TextField {...params} sx={{ height: '50px', width: '150px' }} />}
+              renderInput={params => (
+                <TextField
+                  {...params}
+                  sx={{ height: `${width && width < 330 ? '70px' : '50px'}`, width: '150px' }}
+                />
+              )}
             />
           </LocalizationProvider>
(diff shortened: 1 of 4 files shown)


chaoran-chen (changes requested):
The mobile date range picker view took a hit. https://cov-spectrum-develop-git-comparetobaselinestyle-cov-spectrum.vercel.app/explore/Switzerland/AllSamples/Past6M/variants?variantQuery=nextcladePangoLineage%3ABA.2.75*& looks like this:

image

The public version looks fine:

image

anastasia-escher:
Hi, @chaoran-chen ! Strange, it looks ok on my phone. 
Which browser do you use? 
Screenshot_20220811-174456_Chrome

anastasia-escher:
Ah, your screnshot is for a smaller screen than the one my phone has,  I didn't check it, sorry.

chaoran-chen (approved):
Thanks, this looks good! There is unfortunately now a conflict with the develop branch. Could you resolve it and then merge it, please?

chaoran-chen (changes requested):
Hi @anastasia-escher, I'm sorry but can I request one more change? I think that it is better to move the two small buttons (delete row and move up) to the left side. When I was using the search, I became aware that everything else is on the left side (add variant button, advanced search button, the known variant cards) and that I don't want to move my mouse all the way to the right. I am not sure if others think the same way but my feeling is that it is more comfortable if the buttons are closer together.

Screen Shot 2022-08-17 at 15 53 44

PR #572, opened 2022-08-19
Baseline comparisone style

anastasia-escher:
Change styles for comparison fields.


src/components/VariantSearchField.tsx
```diff
@@ -343,7 +350,11 @@ export const VariantSearchField = ({ onVariantSelect, currentSelection, triggerS
       onDragLeave={dragLeave}
       onDrop={drop}
       className={
-        'p-1 m-1 border-2 border-dashed ' + (dragOngoingDepth ? 'border-black' : 'border-transparent')
+        `${
+          analysisMode === AnalysisMode.CompareToBaseline || analysisMode === AnalysisMode.CompareEquals
```
chaoran-chen: I don't think that the `VariantSearchField` need/should know about the analysis mode. This component is about a single search field. It shouldn't care about for which analysis the selected variant is used and also not about how the field is positioned on the page. That should be the task of the component that uses it.

For the margins in this case, can't you set them instead in `VariantSearch`, e.g., around line 179?

image

chaoran-chen (changes requested):

src/components/VariantSearchField.tsx
```diff
@@ -87,9 +88,15 @@ type Props = {
   onVariantSelect: (selection: VariantSelector) => void;
   isSimple: boolean;
   triggerSearch: () => void;
+  analysisMode?: AnalysisMode;
```
chaoran-chen: After your edits, there is now no need to pass `analysisMode` into `VariantSearchField` anymore, is it?

chaoran-chen (changes requested):

chaoran-chen (approved):
Thanks!

PR #578, opened 2022-08-26
implement dnd

### LINKED ISSUE #555: Drag&drop variants for touch
It would be great if the drag&drop of variants from the known variant list to the variant search (introduced in #551) also works on touch screens. I'm not sure if it would be best to use the [Touch API](https://developer.mozilla.org/en-US/docs/Web/API/Touch) directly or to use a library such as [React DND](https://react-dnd.github.io/react-dnd/docs/overview) that supports a touch backend.

Especially for mobile screens, this could be quite useful.

image

anastasia-escher:
rewrite drag and drop functionality with react-dnd
resolves #555


chaoran-chen:
Thanks @anastasia-escher! The drag and dropping is working very well but unfortunately, I cannot horizontally scroll the variant cards anymore. Could we solve this by going into the drag and drop mode only after the user keeps pressing for a certain amount of time?

chaoran-chen (approved):
Thank you, @anastasia-escher and @dameyerdave! It works great!

PR #604, opened 2022-09-22
change FocusVariantHeaderControls to use mui

### LINKED ISSUE #590: BUG: Dropdown for "Other websites" has gap from trigger button, causing UX issues
I remember this being an issue in the past, but I particularly noticed this during your presentation to SPHERES, where you struggled with this while giving the talk @chaoran-chen 

I can reproduce this problem on Safari on my Mac. But this seems to be a more universal issue given that you had it, too.

anastasia-escher:
fixes #590


chaoran-chen (changes requested):
The view is great but the "other websites" menu is showing too many buttons:

Just compare the [old menu](https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M/variants?nextcladePangoLineage=ba.5*&) and the [new one](https://cov-spectrum-develop-git-otherwebsitesbutton-cov-spectrum.vercel.app/explore/Switzerland/AllSamples/Past6M/variants?nextcladePangoLineage=ba.5*&) for BA.5*:

src/components/FocusVariantHeaderControls.tsx
```diff
@@ -1,23 +1,66 @@
...
+    },
+    '& .MuiMenuItem-root': {
+      '& .MuiSvgIcon-root': {
+        fontSize: 18,
+        color: theme.palette.text.secondary,
+      },
+      '&:active': {
+        backgroundColor: alpha(theme.palette.primary.main, theme.palette.action.selectedOpacity),
+      },
+    },
```
chaoran-chen, line 60: I'm also a bit surprised to see this code. This does not look like the React-style. Are you sure that this is the right way to do it?

src/components/FocusVariantHeaderControls.tsx
```diff
@@ -70,54 +104,106 @@ export const FocusVariantHeaderControls = React.memo(({ selector }: Props): JSX.
...
-          </Button>
-        </ExternalLink>
-
-        <Dropdown.Toggle split variant='secondary' />
-        <Dropdown.Menu>
+      <div style={{ marginRight: '30px', marginTop: '5px' }}>
+        <Button
+          style={{ marginRight: '4px' }}
+          id='demo-customized-button'
+          aria-controls={openSequence ? 'demo-customized-menu' : undefined}
```
chaoran-chen: Do we need/want these lines at all?

src/components/FocusVariantHeaderControls.tsx
```diff
@@ -70,54 +104,106 @@ export const FocusVariantHeaderControls = React.memo(({ selector }: Props): JSX.
...
+          endIcon={<KeyboardArrowDownIcon />}
+        >
+          Sequence
+        </Button>
+        <StyledMenu
+          disableScrollLock={true}
+          id='demo-customized-menu'
+          MenuListProps={{
+            'aria-labelledby': 'demo-customized-button',
+          }}
```
chaoran-chen: Do we need/want these lines at all?

chaoran-chen (commented):
Thanks, @anastasia-escher! I just noticed a few lines of code where I'm not sure whether they are useful. Otherwise, it's great! Feel free to merge after either addressing the comments or right away if you think that the code should stay as it is.

PR #624, opened 2022-10-18
add date range picker to collection page

### LINKED ISSUE #518: Add date range filter to collection page
Currently, the collection page uses data from the past 6 months. The date range should, however, be configurable. The existing date range filter component should be reused and added next to the country filter on the `CollectionSingleViewPage`.

anastasia-escher:
resolves #518

### DIFF
--- a/src/components/DateRangePicker.tsx
+++ b/src/components/DateRangePicker.tsx
@@ -12,12 +12,13 @@ import { useResizeDetector } from 'react-resize-detector';
 
 interface Props {
   dateRangeSelector: DateRangeSelector;
+  setDateRangeSelector?: any;
 }
 
 const minimumDate: Date = new Date('2020-01-06'); // first day of first week of 2020
 const today = new Date();
 
-export const DateRangePicker = ({ dateRangeSelector }: Props) => {
+export const DateRangePicker = ({ dateRangeSelector, setDateRangeSelector }: Props) => {
   const { width, ref } = useResizeDetector<HTMLDivElement>();
 
   const [startDate, setStartDate] = React.useState<Date | null>(null);
@@ -54,6 +55,15 @@ export const DateRangePicker = ({ dateRangeSelector }: Props) => {
           })
         );
       }
+
+      if (setDateRangeSelector) {
+        setDateRangeSelector(
+          new FixedDateRangeSelector({
+            dateFrom: newDateFrom,
+            dateTo: newDateTo,
+          })
+        );
+      }
     }
   };
 
@@ -73,6 +83,7 @@ export const DateRangePicker = ({ dateRangeSelector }: Props) => {
     <>
       <div ref={ref} className='w-full flex flex-row items-center flex-wrap'>
         <HeaderDateRangeSelect exploreUrl={exploreUrl} />
+
         <div className={`flex flex-row ${width && width < 480 ? 'flex-wrap mt-2 mb-2 ml-1' : 'ml-2'}`}>
           <LocalizationProvider dateAdapter={AdapterDateFns}>
             <DatePicker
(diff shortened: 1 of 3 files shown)


chaoran-chen (changes requested):
Hi @anastasia-escher. Could you please also add the "special" date ranges like "past 6 months", etc. (again reusing as much code as possible)? I'm not sure if there is a dedicated component containing the "special" date ranges and the date picker. If not, it would be good to create one and use it both on the normal variant pages and for the collections.

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -63,7 +64,11 @@ export const CollectionSingleViewPage = () => {
       highlightedOnly: queryParams.get('highlightedOnly') === 'true',
     };
   }, [queryString]);
-  const dateRangeSelector = new SpecialDateRangeSelector('Past6M'); // TODO
+
+  // Date range
+  const [dateRangeSelector, setDateRangeSelector] = useState<DateRangeSelector | SpecialDateRangeSelector>(
```
chaoran-chen: `SpecialDateRangeSelector` is a subtype of `DateRangeSelector`, so there should be no need to mention both.

src/components/HeaderDateRangeSelect.tsx
```diff
@@ -39,24 +43,23 @@ export const HeaderDateRangeSelect = ({ exploreUrl }: Props) => {
...
-          Custom Range
-        </option>
-        {specialDateRanges.map(d => (
-          <option value={d} key={specialDateRangeToString(d)}>
-            {specialDateRangeToString(d)}
-          </option>
-        ))}
-      </Form.Control>
-    </Form>
+    <div>
```
chaoran-chen: Why this new `div`?

src/components/HeaderDateRangeSelect.tsx
```diff
@@ -10,23 +10,27 @@ import {
...
 
   const handleChange = (event: React.ChangeEvent<{ value: unknown }>) => {
-    if (isDateRangeEncoded(event.target.value)) {
+    if (isDateRangeEncoded(event.target.value) && exploreUrl) {
       exploreUrl.setDateRange(dateRangeUrlToSelector(event.target.value));
     }
+    if (setDateRangeSelector && isDateRangeEncoded(event.target.value)) {
+      setDateRangeValue(event.target.value);
+      setDateRangeSelector(dateRangeUrlToSelector(event.target.value));
+    }
```
chaoran-chen: It would be nicer to nest the `if`s:

```
if (isDateRangeEncoded(event.target.value)) {
  if (exploreUrl) {...}
  if (setDateRangeSelector) {...}
}
```

src/components/HeaderDateRangeSelect.tsx
```diff
@@ -10,23 +10,27 @@ import {
 
 interface Props {
   exploreUrl?: ExploreUrl;
+  setDateRangeSelector?: any;
```
chaoran-chen: Please propoerly type it instead of using `any`.

src/components/DateRangePicker.tsx
```diff
@@ -12,12 +12,13 @@ import { useResizeDetector } from 'react-resize-detector';
 
 interface Props {
   dateRangeSelector: DateRangeSelector;
+  setDateRangeSelector?: any;
```
chaoran-chen: Please propoerly type it instead of using `any`.

chaoran-chen (changes requested):
Thanks, @anastasia-escher! I think this works fine but the code needs a few minor improvements.

chaoran-chen (approved):
Thanks!

PR #632, opened 2022-10-27
fix index parameter in removeSelection function

### LINKED ISSUE #617: BUG: Bin deletes query in row below, not in same row

anastasia-escher:
fix #617


chaoran-chen (approved):
thanks!

PR #635, opened 2022-11-02
add full names for nexclade

### LINKED ISSUE #586: Show full name of Nextclade Pango lineage
For a pure Pango lineage query that does not use Nextclade, we display the full name:

We should do the same for variants defined by a nextcladePangoLineage.

anastasia-escher:
fix #586


chaoran-chen (approved):
Thanks!

Issue #611, opened 2022-10-03
ENH: Make "now" not "now" but "now - 1 week" in past month etc. periods


corneliusroemer:
The covSpectrum date default selection currently goes till today which has the consequence of showing _very_ noisy periods that should probably better be hidden.

Nothing would be lost by showing data only till "today - 1 week", this would also fix the fact that we often see 0-100% in the number on the right (h/t @emmahodcroft)
<img width="1659" alt="image" src="https://user-images.githubusercontent.com/25161793/193568238-9d8174c9-2bc5-441a-8b42-3a182d2721c2.png">

emmahodcroft:
This was brought up because I always get confused when I have a graph (or something like a graph) but then it says 0% on the right - makes me double-take because I wonder if I've misunderstood if there's no data or some data.

I'd also suggest maybe adding in small font below the "0.0% Proportion 0.0-100.0% confidence int" the total number of sequences in the graph - just lets one do a very quick sanity check to estimate how much data is behind the chart! (One can click on 'absolute' to see this but at a glance it can be hard to see.)

In particular for me it can be tricksy that the confidence interval changes over time, even though there's no data - that's unintuitive for me! 🙃 
![image](https://user-images.githubusercontent.com/14290674/193568763-7bfc4954-8648-4a8b-a1e9-2dc9ca5b2a83.png)

chaoran-chen:
Thanks for the suggestion! I implemented it. I also changed the preview plots on the left side to ignore one week of data.

corneliusroemer:
Thank you so much :)

corneliusroemer:
Excellent, much better! We should think about whether making it 10 days rather than 7 could be better. The most recent data is so unreliable and sparse. But this is easy to change, I can play around with a PR myself ;) 

<img width="1672" alt="image" src="https://user-images.githubusercontent.com/25161793/193998352-ca549b86-adde-40a8-b45d-071bd5f1574c.png">

emmahodcroft:
This is great, thanks @chaoran-chen !

PR #641, opened 2022-11-17
Submission date

### LINKED ISSUE #581: Filter by submission date
Currently, the date filter filters by sampling date. It would be great if there is an option to filter by submission date.

anastasia-escher:
resolves #581 

If 'Clear filter' is clicked, the filter is set to "all times" which is considered to be the default value and is not shown in the url (neither in the advances filter block).

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -50,6 +50,9 @@ export const App = () => {
   const parseDateSubmitted = () => {
     if (dateSubmitted) {
       if (dateSubmitted.dateSubmitted) {
+        if (dateSubmitted.dateSubmitted === 'AllTimes') {
+          return null;
+        }
         return `Submission date: ${dateSubmitted.dateSubmitted}`;
       } else if (dateSubmitted.dateSubmittedFrom && dateSubmitted.dateSubmittedTo) {
         return `Submission date: from ${dateSubmitted.dateSubmittedFrom} to ${dateSubmitted.dateSubmittedTo}`;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -45,14 +45,14 @@ export const App = () => {
   const { width, ref } = useResizeDetector<HTMLDivElement>();
   const isSmallScreen = width !== undefined && width < 768;
 
-  const { host, qc, setHostAndQc, dateSubmittedRaw } = useExploreUrl() ?? {};
+  const { host, qc, setHostAndQc, dateSubmitted } = useExploreUrl() ?? {};
 
   const parseDateSubmitted = () => {
-    if (dateSubmittedRaw) {
-      if (dateSubmittedRaw.dateSubmitted) {
-        return `Submission date: ${dateSubmittedRaw.dateSubmitted}`;
-      } else if (dateSubmittedRaw.dateSubmittedFrom && dateSubmittedRaw.dateSubmittedTo) {
-        return `Submission date: from ${dateSubmittedRaw.dateSubmittedFrom} to ${dateSubmittedRaw.dateSubmittedTo}`;
+    if (dateSubmitted) {
+      if (dateSubmitted.dateSubmitted) {
+        return `Submission date: ${dateSubmitted.dateSubmitted}`;
+      } else if (dateSubmitted.dateSubmittedFrom && dateSubmitted.dateSubmittedTo) {
+        return `Submission date: from ${dateSubmitted.dateSubmittedFrom} to ${dateSubmitted.dateSubmittedTo}`;
       }
     }
     return null;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,7 +52,36 @@ export const App = () => {
       if (dateSubmitted.dateSubmitted === 'AllTimes') {
         return null;
       }
-      return `${dateSubmitted.dateSubmitted}`;
+
+      let dateString: string = '';
+
+      switch (dateSubmitted.dateSubmitted) {
+        case 'Past2W':
+          dateString = 'Past 2 weeks';
+          break;
+        case 'Past1M':
+          dateString = 'Past month';
+          break;
+        case 'Past2M':
+          dateString = 'Past 2 months';
+          break;
+        case 'Past3M':
+          dateString = 'Past 3 months';
+          break;
+        case 'Past6M':
+          dateString = 'Past 6 months';
+          break;
+        case 'Y2020':
+          dateString = '2020';
+          break;
+        case 'Y2021':
+          dateString = '2021';
+          break;
+        case 'Y2022':
+          dateString = '2022';
+      }
+
+      return `${dateString}`;
     } else if (dateSubmitted?.dateSubmittedFrom && dateSubmitted.dateSubmittedTo) {
       return `from ${dateSubmitted.dateSubmittedFrom} to ${dateSubmitted.dateSubmittedTo}`;
     }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,17 +47,16 @@ export const App = () => {
 
   const { host, qc, setHostAndQc, dateSubmitted } = useExploreUrl() ?? {};
 
-  const parseDateSubmitted = () => {
-    if (dateSubmitted) {
-      if (dateSubmitted.dateSubmitted) {
-        if (dateSubmitted.dateSubmitted === 'AllTimes') {
-          return null;
-        }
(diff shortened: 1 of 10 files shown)


chaoran-chen:
Thanks, @anastasia-escher! I took a brief look and it seems to work fine. There is just one problem which I haven't anticipated. Since #611, now is not actually now anymore: e.g., "past 3 months" does not include the full past three months but the past three months minus the most recent 7 days. This makes sense for the sampling date filters (as motivated in #611) but not at all for the submission date filters.

@anastasia-escher, would you know a clean way to make `now=now` for the submission date filter and `now=now-7 days` for the sampling date filters?

src/components/DateRangePicker.tsx
```diff
@@ -33,7 +35,9 @@ export const DateRangePicker = ({
   const [startDate, setStartDate] = React.useState<Date | null>(null);
   const [endDate, setEndDate] = React.useState<Date | null>(null);
 
-  const { dateFrom, dateTo } = dateRangeSelector.getDateRange();
+  const { dateFrom, dateTo } = submission
+    ? dateRangeSelector.getDateRange((submission = true))
+    : dateRangeSelector.getDateRange((submission = false));
```
chaoran-chen: Why not just `dateRangeSelector.getDateRange(submission)`?

src/App.tsx
```diff
@@ -45,7 +45,21 @@ export const App = () => {
...
+      if (dateSubmitted.dateSubmitted) {
+        if (dateSubmitted.dateSubmitted === 'AllTimes') {
+          return null;
+        }
+        return `Submission date: ${dateSubmitted.dateSubmitted}`;
+      } else if (dateSubmitted.dateSubmittedFrom && dateSubmitted.dateSubmittedTo) {
+        return `Submission date: from ${dateSubmitted.dateSubmittedFrom} to ${dateSubmitted.dateSubmittedTo}`;
+      }
+    }
+    return null;
```
chaoran-chen: You can use the optional chaining operator `?.` to reduce the number of if clauses. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/App.tsx
```diff
@@ -45,7 +45,21 @@ export const App = () => {
   const { width, ref } = useResizeDetector<HTMLDivElement>();
   const isSmallScreen = width !== undefined && width < 768;
 
-  const { host, qc, setHostAndQc } = useExploreUrl() ?? {};
+  const { host, qc, setHostAndQc, dateSubmitted } = useExploreUrl() ?? {};
+
+  const parseDateSubmitted = () => {
```
chaoran-chen: Why is it called `parseDateSubmitted`? Would `formatDateSubmittedAsString` be a more appropriate name for this function?

src/App.tsx
```diff
@@ -62,26 +76,33 @@ export const App = () => {
...
+              <div className='flex flex-row'>
+                <FaFilter
+                  className='m-1'
+                  style={{ width: '30px', minWidth: '30px', height: '30px', minHeight: '30px' }}
+                />
+                <div className='ml-4 flex-grow-1'>
+                  <div className='font-weight-bold'>Advanced filters are active</div>
+                  {!isDefaultHostSelector(host) && <div>Selected hosts: {host.join(', ')}</div>}
+                  {!isDefaultQcSelector(qc) && <div>Sequence quality: {formatQcSelectorAsString(qc)}</div>}
+                  {parseDateSubmitted() && <div>{parseDateSubmitted()}</div>}
```
chaoran-chen: To be consistent with the previous code, it would be nice to write `Submission date: ` here and let the function only return the formatted dates.

src/components/DateRangePicker.tsx
```diff
@@ -11,20 +11,33 @@ import { globalDateCache } from '../helpers/date-cache';
 import { useResizeDetector } from 'react-resize-detector';
 
 interface Props {
+  submission?: boolean;
   dateRangeSelector: DateRangeSelector;
   setDateRangeSelector?: React.Dispatch<React.SetStateAction<DateRangeSelector>>;
+  setSubmissionDateRangeSelector?: React.Dispatch<React.SetStateAction<DateRangeSelector>>;
+  setSpecialSubmissionDateRaw?: React.Dispatch<React.SetStateAction<string | null>>;
+  specialSubmissionDateRaw?: string | null;
```
chaoran-chen, line 19: Why do we need to distinguish between the submission date range selector and the regular date range selector here? Why does it make a difference for the DateRangePicker? (I understand the need for the `submission` field but not exactly why we need `setSubmissionDateRangeSelector, setSpecialSubmissionDateRaw, and specialSubmissionDateRaw`.

src/data/DateRangeSelector.ts
```diff
@@ -148,3 +149,74 @@ export function specialDateRangeToString(dateRange: SpecialDateRange): string {
       return '2022';
   }
 }
+
+export interface DateRangeRaw {
```
chaoran-chen, line 153: Could you please explain the purpose of this type? Why are we not transforming directly to/from date range objects?

src/data/DateRangeSelector.ts
```diff
@@ -148,3 +149,74 @@ export function specialDateRangeToString(dateRange: SpecialDateRange): string {
...
+  [key: string]: string | null;
+}
+
+const fields = ['dateSubmittedFrom', 'dateSubmittedTo', 'dateSubmitted'] as const;
```
chaoran-chen: `fields` is not a very clear variable name within this file. These are only fields related to the submission dates but the file is more generic and about date ranges in general.

src/helpers/explore-url.ts
```diff
@@ -132,6 +149,9 @@ export function useExploreUrl(): ExploreUrl | undefined {
     },
     [history, locationState.pathname, locationState.search, routeMatches.locationSamplingDate]
   );
+
+  // http://localhost:3000/explore/Switzerland/AllSamples/Y2020/variants?dateSubmittedFrom=2022-05-09&dateSubmittedTo=2022-11-03&dateSubmittedFrom=2020-01-06&dateSubmittedTo=2022-11-10&nextcladeQcOverallScoreTo=29&dateSubmittedFrom=2022-10-27&dateSubmittedTo=2022-11-03
```
chaoran-chen: Please remove.

src/helpers/explore-url.ts
```diff
@@ -64,6 +75,12 @@ export const defaultAnalysisMode: AnalysisMode = AnalysisMode.Single;
 
 export const defaultHost: HostSelector = [HUMAN];
 
+const deleteSubmissionDateParams = (params: URLSearchParams) => {
```
chaoran-chen: I think that this should be moved to `DateRangeSelector` where also the other URLSearchParams-manipulating functions for date ranges are located.

chaoran-chen (changes requested):

chaoran-chen (changes requested):
Hi @anastasia-escher. I found an issue: The submission date filters get automatically set. To reproduce: 

1. Open the [landing page](https://cov-spectrum-develop-git-submissiondate-cov-spectrum.vercel.app/)
2. Click on the "advanced filter" button
3. Click on "Save"

The submission date filter is active although we did not explicitly set it.

chaoran-chen (changes requested):
Another thing: it would be better to show the formatted strings and not the internal values: i.e., "Past 3 months" instead of "Past3M"

image

chaoran-chen (changes requested):
The default form doesn't make sense. By default (if no filter is set), it's equivalent to "all times" and not to "past 6 months". Thus, it should also say "all times". And if the user clicks on "clear filter", it should reset to "all times".

image

src/App.tsx
```diff
@@ -45,7 +45,49 @@ export const App = () => {
...
+
+  const formatDateSubmittedAsString = () => {
+    if (dateSubmitted?.dateSubmitted) {
+      if (dateSubmitted.dateSubmitted === 'AllTimes') {
+        return null;
+      }
+
+      let dateString: string = '';
+
+      switch (dateSubmitted.dateSubmitted) {
```
chaoran-chen, line 58: Please reuse the [existing function](https://github.com/cevo-public/cov-spectrum-website/blob/3c18131ead880099d4ae3344f80bd58003328466/src/data/DateRangeSelector.ts#L129)

chaoran-chen (changes requested):

PR #644, opened 2022-11-25
set 'Submitted in past 10 days' field type as numeric; the same for '…

### LINKED ISSUE #598: ENH: Allow numeric filtering in collections
Collections are amazing. One thing that would make them even more useful would be if I could filter numerically:
e.g. select only lineages with at least 10 submissions in the past 10 day. Right now I don't think that's possible as filtering assumes that the columns are strings and not numbers:
image

This could be as simple as adding a line to the relevant table column settings (if we're lucky)

anastasia-escher:
…Number sequences'
solves #598


chaoran-chen (approved):
Nice! Thanks!

PR #650, opened 2022-11-30
Mutation overview

### LINKED ISSUE #607: Mutation overview in collections
**Aim:** It should be easier to get an overview of the mutations of the variants in a collection.

**Idea:** A new tab "mutations" should be added to the collection pages. It should contain a table with the following columns if no baseline is selected:

- \>90% (which contains fixed mutations/mutations that occur in more than 90% of the sequences)
- 60-90%
- 30-60%
- 5-30%
- "See all"-button that links to the mutations box on the variant page

If a baseline is selected, the table will compare the mutation set of the variants in the collection to the **fixed/>90%** mutations of the baseline. It shows a nested header like this:

image

anastasia-escher:
Resolves #607

### DIFF
--- a/src/components/SequencesOverTimeTabContent.tsx
+++ b/src/components/SequencesOverTimeTabContent.tsx
@@ -1,0 +1,103 @@
+import { useRef } from 'react';
+import { Collection } from '../data/Collection';
+import { Dataset } from '../data/Dataset';
+import { LocationDateVariantSelector } from '../data/LocationDateVariantSelector';
+import { PprettyRequest } from '../data/ppretty/ppretty-request';
+import { DateCountSampleEntry } from '../data/sample/DateCountSampleEntry';
+import { formatVariantDisplayName, VariantSelector } from '../data/VariantSelector';
+import { VariantTimeDistributionChartWidget } from '../widgets/VariantTimeDistributionChartWidget';
+import { PprettyGridExportButton } from './CombinedExport/PprettyGridExportButton';
+import {
+  PprettyGridExportManager,
+  PprettyGridExportManagerContext,
+} from './CombinedExport/PprettyGridExportManager';
+import { GridCell, PackedGrid } from './PackedGrid';
+import { ErrorAlert } from './ErrorAlert';
+
+type SequencesOverTimeTabContentProps = {
+  collection: Collection;
+  variants: { query: VariantSelector; name: string; description: string }[];
+  variantsDateCounts: PromiseSettledResult<Dataset<LocationDateVariantSelector, DateCountSampleEntry[]>>[];
+  baselineDateCounts: Dataset<LocationDateVariantSelector, DateCountSampleEntry[]>;
+  mode: 'Single' | 'CompareToBaseline';
+};
+
+export default function SequencesOverTimeTabContent({
+  collection,
+  variants,
+  variantsDateCounts,
+  baselineDateCounts,
+  mode,
+}: SequencesOverTimeTabContentProps) {
+  const exportManagerRef = useRef(
+    new PprettyGridExportManager(requests => {
+      // TODO What happens if the arguments of SequencesOverTimeTabContent change? We probably need to update this
+      //   function as well?
+      const variantToDetailsMap = new Map<string, { name: string; description: string }>();
+      for (let { query, name, description } of collection.variants) {
+        const variantString = formatVariantDisplayName(JSON.parse(query));
+        variantToDetailsMap.set(variantString, { name, description });
+      }
+      const mergedRequest: PprettyRequest = {
+        config: {
+          plotName: 'sequences-over-time_collection',
+          plotType: 'line',
+          sizeMultiplier: Math.max(Math.ceil(Math.sqrt(collection.variants.length)) / 2, 1),
+        },
+        metadata: {
+          location: requests[0].metadata.location,
+          collection: {
+            id: collection.id,
+            title: collection.title,
+            maintainer: collection.maintainers,
+          },
+        },
+        data: [],
+      };
+      for (let request of requests) {
+        mergedRequest.data.push(
+          ...request.data.map((d: any) => {
+            const variant = request.metadata.variant;
+            const { name, description } = variantToDetailsMap.get(variant)!;
+            return {
+              ...d,
+              variant,
+              name,
+              description,
+            };
+          })
+        );
+      }
+      return mergedRequest;
+    })
+  );
+
+  return (
+    <PprettyGridExportManagerContext.Provider value={exportManagerRef.current}>
+      <PprettyGridExportButton />
+      <PackedGrid maxColumns={3}>
+        {variants.map((variant, i) => {
+          const vdc = variantsDateCounts[i];
+          if (vdc.status === 'fulfilled') {
+            return (
+              <GridCell minWidth={600} key={i}>
+                <VariantTimeDistributionChartWidget.ShareableComponent
+                  title={mode === 'Single' ? variant.name : `Comparing ${variant.name} to baseline`}
+                  height={350}
+                  variantSampleSet={vdc.value}
+                  wholeSampleSet={baselineDateCounts}
+                />
+              </GridCell>
+            );
+          } else {
+            return (
+              <GridCell minWidth={600} key={i}>
+                <ErrorAlert messages={[vdc.reason.message]} />
+              </GridCell>
+            );
+          }
+        })}
+      </PackedGrid>
+    </PprettyGridExportManagerContext.Provider>
+  );
+}
--- a/src/components/TableTabContent.tsx
+++ b/src/components/TableTabContent.tsx
@@ -1,0 +1,263 @@
+import { DataGrid, GridColDef, GridComparatorFn, GridRenderCellParams } from '@mui/x-data-grid';
+import { useEffect, useState, useMemo } from 'react';
+import { AiFillStar } from 'react-icons/ai';
+import { Dataset } from '../data/Dataset';
+import { DateRangeSelector } from '../data/DateRangeSelector';
+import { dateRangeUrlFromSelector } from '../data/DateRangeUrlEncoded';
+import { LocationDateVariantSelector } from '../data/LocationDateVariantSelector';
+import { encodeLocationSelectorToSingleString, LocationSelector } from '../data/LocationSelector';
+import { DateCountSampleEntry } from '../data/sample/DateCountSampleEntry';
+import {
+  addVariantSelectorToUrlSearchParams,
+  formatVariantDisplayName,
+  VariantSelector,
+} from '../data/VariantSelector';
+import { PromiseQueue } from '../helpers/PromiseQueue';
+import { Chen2021FitnessResponse, ValueWithCI } from '../models/chen2021Fitness/chen2021Fitness-types';
+import { getModelData } from '../models/chen2021Fitness/loading';
+import { HashLink as Link } from 'react-router-hash-link';
+import Loader from '../components/Loader';
+import { Button, ButtonVariant } from '../helpers/ui';
+import { csvStringify } from '../helpers/csvStringifyHelper';
+import download from 'downloadjs';
+
+type TableTabContentProps = {
+  collectionId: number;
+  locationSelector: LocationSelector;
+  variants: { query: VariantSelector; name: string; description: string }[];
+  variantsDateCounts: PromiseSettledResult<Dataset<LocationDateVariantSelector, DateCountSampleEntry[]>>[];
+  variantsNumberNewSequences: PromiseSettledResult<number>[];
+  allWholeDateCounts: PromiseSettledResult<Dataset<LocationDateVariantSelector, DateCountSampleEntry[]>>[];
+  dateRangeSelector: DateRangeSelector;
+};
+
+export default function TableTabContent({
+  collectionId,
+  locationSelector,
+  variants,
+  variantsDateCounts,
+  variantsNumberNewSequences,
+  allWholeDateCounts,
+  dateRangeSelector,
+}: TableTabContentProps) {
+  // Fetch relative growth advantage
+  const [relativeAdvantages, setRelativeAdvantages] = useState<
+    (Chen2021FitnessResponse | undefined | 'failed')[]
+  >([]);
+  useEffect(() => {
+    const fetchQueue = new PromiseQueue(10);
+    for (let i = 0; i < variantsDateCounts.length; i++) {
+      const variantDateCountsStatus = variantsDateCounts[i];
+      const allWholeDateCountsStatus = allWholeDateCounts[i];
+      const setFailedFunc = (prev: (Chen2021FitnessResponse | undefined | 'failed')[]) => {
+        const newArr = [...prev];
+        newArr[i] = 'failed';
+        return newArr;
+      };
+      if (variantDateCountsStatus.status === 'rejected' || allWholeDateCountsStatus.status === 'rejected') {
+        fetchQueue.addTask(() => {
+          return new Promise<void>(resolve => {
+            setRelativeAdvantages(setFailedFunc);
+            resolve();
+          });
+        });
+        continue;
+      }
+      const variantDateCounts = variantDateCountsStatus.value;
+      const wholeDateCounts = allWholeDateCountsStatus.value;
+      const totalSequences = variantDateCounts.payload.reduce((prev, curr) => prev + curr.count, 0);
+      fetchQueue.addTask(() => {
+        if (totalSequences > 0) {
+          return getModelData(variantDateCounts, wholeDateCounts, { generationTime: 7 }).then(
+            ({ response }) => {
+              if (response === undefined) {
+                setRelativeAdvantages(setFailedFunc);
+              } else {
+                setRelativeAdvantages(prev => {
+                  const newArr = [...prev];
+                  newArr[i] = response;
+                  return newArr;
+                });
+              }
+            }
+          );
+        } else {
+          // No need to calculate the advantage if there is no available sequence
+          return new Promise<void>(resolve => {
+            setRelativeAdvantages(setFailedFunc);
+            resolve();
+          });
+        }
+      });
+    }
+  }, [variantsDateCounts, allWholeDateCounts, setRelativeAdvantages]);
+
+  // Table definition and data
+  const tableColumns: GridColDef[] = [
+    {
+      field: 'highlighted',
+      headerName: '',
+      width: 40,
+      minWidth: 40,
+      renderCell: (params: GridRenderCellParams<string>) => {
+        const highlighted = params.row.highlighted;
+        return highlighted ? <AiFillStar size='1.5em' className='text-yellow-400' /> : <></>;
+      },
+    },
+    {
+      field: 'name',
+      headerName: 'Name',
+      minWidth: 200,
+      renderCell: (params: GridRenderCellParams<string>) => {
+        const query = params.row.query;
+        const urlParams = new URLSearchParams();
+        addVariantSelectorToUrlSearchParams(query, urlParams);
+
+        const placeString = encodeLocationSelectorToSingleString(locationSelector);
+        return params.value ? (
+          <Link
+            to={`/explore/${placeString}/AllSamples/${dateRangeUrlFromSelector(
+              dateRangeSelector
+            )}/variants?${urlParams.toString()}`}
+            className='overflow-hidden'
+          >
+            <button className='underline break-words overflow-hidden w-full'>{params.value}</button>
+          </Link>
+        ) : (
+          <></>
+        );
+      },
+    },
+    {
+      field: 'queryFormatted',
+      headerName: 'Query',
+      minWidth: 300,
+      renderCell: (params: GridRenderCellParams<string>) => {
+        return <span className='break-words overflow-hidden'>{params.value}</span>;
+      },
+    },
+    { field: 'total', headerName: 'Number sequences', minWidth: 150, type: 'number' },
+    { field: 'newSequences', headerName: 'Submitted in past 10 days', minWidth: 200, type: 'number' },
+    {
+      field: 'advantage',
+      headerName: 'Relative growth advantage',
+      minWidth: 200,
+      sortComparator: sortRelativeGrowthAdvantageValues,
+    },
+    {
+      field: 'advantageCiLower',
+      headerName: 'CI (low)',
+      minWidth: 100,
+      sortComparator: sortRelativeGrowthAdvantageValues,
+    },
+    {
+      field: 'advantageCiUpper',
+      headerName: 'CI (high)',
+      minWidth: 100,
+      sortComparator: sortRelativeGrowthAdvantageValues,
+    },
+    { field: 'description', headerName: 'Description', minWidth: 450 },
+  ];
+
+  const variantTableData = useMemo(() => {
+    return variants.map((variant, i) => {
+      let advantage = undefined;
+      let errorMessage: string | undefined = undefined;
+      const relativeAdvantage: Chen2021FitnessResponse | 'failed' | undefined = relativeAdvantages[i];
+      if (relativeAdvantage === 'failed') {
+        errorMessage = "Can't be calculated";
+      } else if (relativeAdvantage === undefined) {
+        errorMessage = 'Calculating...';
+      } else {
+        advantage = relativeAdvantage.params.fd;
+      }
+      const vcd = variantsDateCounts[i];
+      const vnns = variantsNumberNewSequences[i];
+      if (vcd.status === 'fulfilled') {
+        if (vnns.status !== 'fulfilled') {
+          // Strange that one request was successful but the other one wasn't
+          throw new Error('Unexpected error');
+        }
+        return {
+          id: i,
+          ...variant,
+          name: variant.name.length > 0 ? variant.name : formatVariantDisplayName(variant.query),
+          queryFormatted: formatVariantDisplayName(variant.query),
+          total: vcd.value.payload.reduce((prev, curr) => prev + curr.count, 0),
+          newSequences: vnns.value,
+          advantage: errorMessage ?? ((advantage as ValueWithCI).value * 100).toFixed(2) + '%',
+          advantageCiLower: errorMessage
+            ? '...'
+            : ((advantage as ValueWithCI).ciLower * 100).toFixed(2) + '%',
+          advantageCiUpper: errorMessage
+            ? '...'
+            : ((advantage as ValueWithCI).ciUpper * 100).toFixed(2) + '%',
+        };
+      } else {
+        return {
+          id: i,
+          ...variant,
+          name: variant.name.length > 0 ? variant.name : formatVariantDisplayName(variant.query),
+          queryFormatted: formatVariantDisplayName(variant.query),
+          total: vcd.reason,
+          newSequences: '-',
+          advantage: '-',
+          advantageCiLower: '-',
+          advantageCiUpper: '-',
+        };
+      }
+    });
+  }, [variants, variantsDateCounts, variantsNumberNewSequences, relativeAdvantages]);
+
+  const downloadAsCsv = () => {
+    const csvData = variantTableData.map(x => ({
+      name: x.name,
+      query: x.queryFormatted,
+      number_sequences: x.total,
+      submitted_past_10_days: x.newSequences,
+      relative_growth_advantage: x.advantage,
+      relative_growth_advantage_low: x.advantageCiLower,
+      relative_growth_advantage_high: x.advantageCiUpper,
+      description: x.description,
+    }));
+    download(csvStringify(csvData), `collection-${collectionId}.csv`, 'text/csv');
+  };
+
+  return (
+    <>
+      {variantTableData ? (
+        <>
+          <Button variant={ButtonVariant.PRIMARY} className='w-40' onClick={downloadAsCsv}>
+            Download CSV
+          </Button>
+          <div className='mt-4'>
+            <DataGrid
+              columns={tableColumns}
+              rows={variantTableData}
+              autoHeight={true}
+              getRowHeight={() => 'auto'}
+              density={'compact'}
+            />
+          </div>
+        </>
+      ) : (
+        <Loader />
+      )}
+    </>
+  );
+}
+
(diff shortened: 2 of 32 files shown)


src/pages/CollectionSingleViewPage.tsx
```diff
@@ -359,6 +400,144 @@ const TabPanel = ({ children, value, index, ...other }: TabPanelProps) => {
...
+
+        console.log('baseline', data);
+
+        return data;
+      }
+    },
+    [baselineVariant, mode, locationSelector, dateRangeSelector]
+  );
+
+  // #####################################
```
fengelniederhammer: How about extracting some of those hook usages in custom hooks? That would make the function more readable and we don't need this visual separator in the code.

fengelniederhammer:
Maybe add a test for the new components? :)

src/data/MutationsOnCollectionPage.ts
```diff
@@ -0,0 +1,32 @@
+import { MutationProportionEntry } from './MutationProportionEntry';
+import { VariantSelector } from './VariantSelector';
+
```
chaoran-chen: File names starting with a capital letter usually indicate that it holds a component, type, or class of the same name. That's not the case here.

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,179 @@
...
+      </div>
+      <br />
+      <div style={{ display: 'block', color: 'blue', margin: '10px' }}>
+        <button className='underline break-words overflow-hidden w-full'>
+          <Link
+            to={`/explore/${placeString}/AllSamples/Past6M/variants?${urlParams.toString()}#mutations`}
+            className='overflow-hidden'
+          >
+            See all
+          </Link>
```
chaoran-chen: 1. This can be moved into the above link. We don't need a separate link.
2. It should also respect the date filter that we have on the page.
3. When a baseline is selected, it should open the compare page (not "compare to baseline" because the "compare to base line" mode at the moment does not show the mutation differences)

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -359,6 +400,144 @@ const TabPanel = ({ children, value, index, ...other }: TabPanelProps) => {
...
+    }
+  }, [mutations]);
+
+  const baselineTableData = useMemo(() => {
+    if (mode === 'CompareToBaseline' && baselineMutations.data && mutations.data) {
+      const data: BaselineData = {
+        nucMutations: baselineMutations.data.nucMutations,
+        aaMutations: baselineMutations.data.aaMutations,
+      };
+      console.log('baselineMutations', baselineMutations);
```
chaoran-chen: Please remove logging

src/pages/CollectionSingleViewPage.tsx
chaoran-chen: I like the idea!

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -359,6 +400,144 @@ const TabPanel = ({ children, value, index, ...other }: TabPanelProps) => {
...
+        const baselineData = Promise.all([
+          fetchSamplesCount(selector, signal),
+          MutationProportionData.fromApi(selector, 'aa', signal),
+          MutationProportionData.fromApi(selector, 'nuc', signal),
+        ]).then(async ([variantCount, aaMutationDataset, nucMutationDataset]) => {
+          data.aaMutations = aaMutationDataset.payload;
+          data.nucMutations = nucMutationDataset.payload;
+        });
+
+        console.log('baseline', data);
```
chaoran-chen: Please remove logging

chaoran-chen (changes requested):
Sorry that it took me so long to review!

Is the compare-to-baseline mode working correctly? From #607: "If a baseline is selected, the table will compare the mutation set of the variants in the collection to the fixed/>90% mutations of the baseline." - Where are the mutations of the baseline filtered?

Comparing [BF.7.14 on this collection page](https://cov-spectrum-develop-git-mutationoverview-cov-spectrum.vercel.app/collections/136) and the [Venn diagrams on the compare variants](https://cov-spectrum-develop-git-mutationoverview-cov-spectrum.vercel.app/collections/136), if the Venn diagram is set to 90%, shouldn't we see the same values in the "Additional >90%" column and in the "BF.7 [...]"-only part of the Venn diagram?

image

image

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,171 @@
...
+              mutations.map(item => (
+                <tr key={item.name}>
+                  <td>{renderNameCell(item.query, locationSelector, item.name)}</td>
+                  {mutationColumnsProperties.map(field => {
+                    return (
+                      <td>
+                        {mutationType === 'aa'
+                          ? pullAll(
+                              getNames(filterMutations(item.aaMutations, field)),
+                              getNames(filterMutations(baselineData.aaMutations, field))
```
chaoran-chen: Isn't this wrong? Isn't this comparing e.g., 60-90% of a variant to 60-90% of the baseline instead of always comparing to >90% as stated in the issue:

> If a baseline is selected, the table will compare the mutation set of the variants in the collection to the fixed/>90% mutations of the baseline.

It would actually be great to add this as an explanation to the table just above the table.

chaoran-chen (changes requested):
In addition to the above comment, I see the following tasks remaining:

- When clicking on the name of a variant:
  - [x] The date filter should be the same
  - [x] If a baseline is selected, it should open the compare page (not "compare to baseline" because the "compare to baseline" mode at the moment does not show the mutation differences)
- [x] Clean up the code and remove the console.logs
- [x] Sort the mutations as usual

@fengelniederhammer @JonasKellerer, you are also welcome to review as well if you like. It's a lot of code and having a few more people looking over it can be helpful.

src/components/__tests__/MutationTables.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+        <BaselineMutationsTable
+          baselineData={baselineData}
+          mutations={mutations}
+          mutationType='aa'
+          locationSelector={locationSelector}
+        />
+      </BrowserRouter>
+    );
+
+    const table = getByTestId('baseline-mutation-table');
```
fengelniederhammer: IntelliJ complains: "ESLint: Avoid destructuring queries from `render` result, use `screen.getByTestId` instead(testing-library/prefer-screen-queries)"

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+export type BaselineMutationsTableProps = {
+  baselineData: BaselineData;
+  mutations: MutationObject[] | undefined;
+  mutationType: string;
+  locationSelector: LocationSelector;
+  baselineVariant?: VariantSelector;
+  dateRangeSelector: DateRangeSelector;
+};
+
+export const BaselineMutationsTable = ({
```
fengelniederhammer: The test complains:
```  
console.error
    Warning: Each child in a list should have a unique "key" prop.
    
    Check the render method of `BaselineMutationsTable`. See https://reactjs.org/link/warning-keys for more information.
```

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+  baselineData,
+  mutations,
+  mutationType,
+  locationSelector,
+  baselineVariant,
+  dateRangeSelector,
+}: BaselineMutationsTableProps) => {
+  const headers: string[] = ['Additional', 'Missing'];
+  console.log('baseline mutations', baselineData);
```
fengelniederhammer: Please remove this :)

src/data/mutationsOnCollectionPageData.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface MutationColumnProperty {
+  min: number;
+  max: number;
+  label: string;
+}
+
+export interface MutationObject {
+  query: VariantSelector;
+  name: string;
+  variantCout: number;
```
fengelniederhammer: There is a typo: `variantCouNt`

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -359,13 +402,149 @@ const TabPanel = ({ children, value, index, ...other }: TabPanelProps) => {
...
+          variant: baselineVariant,
+          samplingStrategy: SamplingStrategy.AllSamples,
+          dateRange: dateRangeSelector,
+        });
+
+        Promise.all([
+          fetchSamplesCount(selector, signal),
+          MutationProportionData.fromApi(selector, 'aa', signal),
+          MutationProportionData.fromApi(selector, 'nuc', signal),
+        ]).then(async ([variantCount, aaMutationDataset, nucMutationDataset]) => {
```
fengelniederhammer: `variantCount` is unused -> Please either use it or delete the unnecessary call.

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -327,6 +343,33 @@ export const CollectionSingleViewPage = () => {
               <Loader />
             )}
           </TabPanel>
+          <TabPanel value={tab} index={2}>
+            <FormControl>
+              <RadioGroup
+                aria-labelledby='demo-radio-buttons-group-label'
+                defaultValue='female'
```
fengelniederhammer: This is probably a left over from copy-pasting? :)

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -61,6 +68,7 @@ export const CollectionSingleViewPage = () => {
   const history = useHistory();
   const locationState = useLocation();
   const [tab, setTab] = useState(0);
+  const [mutationType, setMutationType] = useState('aa');
```
fengelniederhammer: Can we narrow the type to `useState<'aa' | 'nuc'>`?

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -359,13 +402,149 @@ const TabPanel = ({ children, value, index, ...other }: TabPanelProps) => {
...
+          <div className='mt-4'>
+            <SingleMutationsTable
+              mutations={mutationTableData}
+              locationSelector={locationSelector}
+              mutationType={mutationType}
+              dateRangeSelector={dateRangeSelector}
+            />
+          </div>
+        </>
+      ) : baselineTableData && mode === 'CompareToBaseline' ? (
```
fengelniederhammer: Why is it not necessary to check for `mutationTableData` here? I would have expected the check to be `mutationTableData && baselineTableData && mode === 'CompareToBaseline'`

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+export const SingleMutationsTable = ({
+  mutations,
+  locationSelector,
+  mutationType,
+  dateRangeSelector,
+}: SingleMutationsTableProps) => {
+  return (
+    <>
+      {mutations && (
```
fengelniederhammer: Why do we do the same check twice? The parent component already checks for truthiness of `mutations`

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+import { sortMutationList } from '../helpers/mutation';
+
+type SingleMutationsTableProps = {
+  mutations: MutationObject[] | undefined;
+  locationSelector: LocationSelector;
+  mutationType: string;
+  dateRangeSelector: DateRangeSelector;
+};
+
+const renderNameCell = (
```
fengelniederhammer: This is a component. Component names should start with a capital letter. Maybe simply rename this to `NameCell`? Also please use it like a component (props as argument, call it via JSX).

src/pages/CollectionSingleViewPage.tsx
fengelniederhammer: There is even already the `SequenceType` for that.

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -359,13 +402,149 @@ const TabPanel = ({ children, value, index, ...other }: TabPanelProps) => {
...
+        })
+      );
+
+      return data;
+    },
+    [variants, locationSelector, dateRangeSelector]
+  );
+
+  const baselineMutations = useQuery(
+    async signal => {
```
fengelniederhammer: The problem, why sometimes data doesn't load, is: This async function isn't actually async, since it does not `await` anything. 

How to reproduce it consistently:
* choose a collection
* select a baseline
* switch to the "mutations" tab

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

"Top-level code, up to and including the first await expression (if there is one), is run synchronously. In this way, an async function without an await expression will run synchronously."
-> The function as it is now will initialize an empty data object, construct a promise and immediately return the empty data. 

I'd propose to restructure this function to use the promises in a more idiomatic way:
```
  const baselineMutations = useQuery(
    signal => {
      if (mode !== 'CompareToBaseline') {
        return Promise.resolve({
          aaMutations: [],
          nucMutations: [],
        });
      }

      const selector: LapisSelector = addDefaultHostAndQc({
        location: locationSelector,
        variant: baselineVariant,
        samplingStrategy: SamplingStrategy.AllSamples,
        dateRange: dateRangeSelector,
      });

      return Promise.all([
        MutationProportionData.fromApi(selector, 'aa', signal),
        MutationProportionData.fromApi(selector, 'nuc', signal),
      ]).then(async ([aaMutationDataset, nucMutationDataset]) => {
        return {
          aaMutations: aaMutationDataset.payload,
          nucMutations: nucMutationDataset.payload,
        };
      });
    },
    [baselineVariant, mode, locationSelector, dateRangeSelector]
  );
```

The same applies to the function in the "mutations useQuery". The usage of promises mirrors more closely the flow of the data instead of setting data via side effects. (Also notice that the function doesn't need to be explicitly `async` because it returns a Promise anyway.)

The problem here can also be fixed by simply `await`ing the Promise in the async function.

fengelniederhammer (changes requested):

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+                              )
+                            ).join(', ')}
+                      </td>
+                    );
+                  })}
+
+                  {mutationColumnsProperties.map(field => {
+                    return (
+                      <td key={field.min}>
+                        {mutationType === 'aa'
```
fengelniederhammer: There is a lot of duplication in those places. It always follows the same pattern:
`mutationType === 'aa' ? doSomething(item.aaMutations) : doSomething(item.nucMutations)`

So the code does exactly the same (as far as I can judge), except for "either get aa or nuc mutations" and "sort according to aa or nuc".

Could we generalize the code a bit to reduce duplication? Introduce a clcass that wraps around mutationType and the mutation data? But there is no need to do that if it's too much work.

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -359,13 +402,149 @@ const TabPanel = ({ children, value, index, ...other }: TabPanelProps) => {
...
+type MutationTableTabContentProps = {
+  mutationType: string;
+  locationSelector: LocationSelector;
+  variants: { query: VariantSelector; name: string; description: string }[];
+  dateRangeSelector: DateRangeSelector;
+  mode: 'Single' | 'CompareToBaseline';
+  baselineVariant?: VariantSelector;
+};
+
+const MutationTableTabContent = ({
```
fengelniederhammer: Can we maybe cut this component in smaller pieces? I'd propose the following structure (in pseudocode):

```
function MutationTableTabContent (props) {
  return mode === 'Single' ? <SingleMutationTableTabContent /> : <BaselineMutationTableTabContent  />
} 

function SingleMutationTableTabContent() {
  data = useMutationsData(selectors, variants, ...)
  return <SingleMutationsTable data={data} />
}

function BaselineMutationTableTabContent() {
  data = useBaselineMutationsData(selectors, variants, ...)
  return <BaselineMutationsTable data={data}
}

function useBaselineMutationsData(...) {
  mutationsData = useMutationsData(selectors, variants, ...)

  // other useQuery, useMemo etc. for baseline data only

  return { mutationsData , baselineData }
}

function useMutationsData(...) {
  // useQuery, useMemo etc.
  return mutationsData 
}
```

Maybe even in a separate file? The current file grew to 800 lines which is quite a lot.

src/components/__tests__/MutationTables.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+      nucMutations: [
+        { proportion: 0.051, mutation: 'C20404T', count: 4 },
+        { proportion: 0.091, mutation: 'T670G', count: 5 },
+      ],
+    },
+  ];
+
+  it('the table is rendered', () => {
+    render(
+      <BrowserRouter>
```
fengelniederhammer: Doesn't change much here, but I think it's better to use `MemoryRouter` in tests: https://v5.reactrouter.com/web/guides/testing

src/components/VariantMutations.tsx
```diff
@@ -259,7 +259,7 @@ export const VariantMutations = ({ selector }: Props) => {
 
   return (
     <>
-      <div className='mb-8'>
+      <div className='mb-8' id='mutations'>
```
fengelniederhammer: Why do we need this id?

src/pages/FocusSinglePage.tsx
```diff
@@ -511,7 +511,7 @@ export const FocusSinglePageContent = ({
               </Sentry.ErrorBoundary>
             </div>
 
-            <div className='m-4'>
+            <div className='m-4' id='mutations1'>
```
fengelniederhammer: Why do we need this id?

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -448,10 +628,13 @@ const TableTabContent = ({
         const query = params.row.query;
         const urlParams = new URLSearchParams();
         addVariantSelectorToUrlSearchParams(query, urlParams);
+
         const placeString = encodeLocationSelectorToSingleString(locationSelector);
         return params.value ? (
           <Link
-            to={`/explore/${placeString}/AllSamples/Past6M/variants?${urlParams.toString()}`}
+            to={`/explore/${placeString}/AllSamples/${dateRangeUrlFromSelector(
+              dateRangeSelector
```
fengelniederhammer: Why do we need the dateRangeSelector here now? Was this a bug before?

src/pages/CollectionSingleViewPage.tsx
fengelniederhammer: If we wrap the `FormControl` and the `MutationTableTabContent` in another component, we can move this state out of this far-too-big component into that other component.

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+  );
+};
+
+const getNames = (mutations: MutationProportionEntry[]) => {
+  return mutations.map(item => item.mutation);
+};
+
+function commonElements<T>(arr1: T[], arr2: T[]): T[] {
+  return arr1.filter(function (element) {
+    return arr2.indexOf(element) !== -1;
```
fengelniederhammer: Out of interest: Is `indexOf` better than `includes`? `includes` would make the intention clearer.

fengelniederhammer: ~I think one could even shorten this to `return arr1.filter(arr2.includes);` then.~ Somehow doesn't work.

src/pages/CollectionSingleViewPage.tsx
fengelniederhammer: The downside is that we lose caching that way, because the self-written useQuery doesn't have a cache. However the browser seems to have some basic caching there, so we might save some requests.

Using `react-query` would really help here, because it has a proper built-in cache.

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+                    );
+                  })}
+
+                  {mutationColumnsProperties.map(field => {
+                    return (
+                      <td key={field.min}>
+                        {mutationType === 'aa'
+                          ? sortMutationList(
+                              'aa',
+                              compareToBaseline(
```
fengelniederhammer: What is this supposed to do? I don't quite get the meaning of "Missing" yet. Missing from what? What do the percentages mean (90% of what)?

The code selects mutations that are in the baseline and the complement of `field.min` in `item`. How does this connect to the headline "Missing"?

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,257 @@
...
+                      true,
+                      dateRangeSelector,
+                      baselineVariant
+                    )}
+                  </td>
+                  {mutationColumnsProperties.map(field => {
+                    return (
+                      <td key={field.label}>
+                        {mutationType === 'aa'
+                          ? sortMutationList(
```
fengelniederhammer: This component contains both: display of data and the logic. I would prefer to separate both for testability:
* The logic: Given some mutation data, returns data to display -> Test the data selection logic (Given mutation and baseline data, some mutation ends up in the xx% segment of the table.)
* The display: Given the output of the previous function, it renders it as a table. -> Test display logic only (is a nice table).

src/components/MutationsTable.tsx
chaoran-chen: `arr1.filter(arr2.includes)` looks dangerous because `filter` does not just pass the element to the callback function but actually `(element, index, array)`, and `includes` accepts `fromIndex` as its second argument. I.e., `filter's index` will be used as `includes' fromIndex`

src/components/MutationsTable.tsx
anastasia-escher: The name of the function was a bit misleading, so I've just renamed it to 'findMissingMutations'. It should find missing mutations based on the following logic (as explained by @chaoran-chen): "If a fixed mutation of the baseline is present in 25% of variant samples - in other words missing in 75% of the samples, it should occur in the column 60-90% but in none of the others."

src/pages/CollectionSingleViewPage.tsx
anastasia-escher: Yes, the link should open the main page with the date selector specified on the collection page.

src/pages/FocusSinglePage.tsx
anastasia-escher: The link to the main page from the Mutation Tab should open the main page on the section with mutations, so this id was added in order to use Hashlink.

src/components/VariantMutations.tsx
anastasia-escher: The link form the mutations tab should lead to the mutations section on the main page, so we need it in order to use a hash link.

fengelniederhammer: It works, but only if the page is already fully loaded. When the page is still building up, I end up in the wrong place on the page.

src/components/MutationsTable.tsx
anastasia-escher: Without this check I get the "the object is probably undefined" error in mutations.map(...).

fengelniederhammer: You can require in the Props type that it is a `MutationObject[]` instead of `MutationObject[] | undefined`

src/components/MutationTableTabContent.tsx
```diff
@@ -66,31 +66,30 @@ export default function MutationTableTabContent({
...
+        samplingStrategy: SamplingStrategy.AllSamples,
+        dateRange: dateRangeSelector,
+      });
 
-        return data;
-      }
+      return Promise.all([
+        MutationProportionData.fromApi(selector, 'aa', signal),
+        MutationProportionData.fromApi(selector, 'nuc', signal),
+      ]).then(async ([aaMutationDataset, nucMutationDataset]) => {
```
fengelniederhammer: This doesn't need to be async.

src/pages/CollectionSingleViewPage.tsx
```diff
@@ -325,7 +325,6 @@ export const CollectionSingleViewPage = () => {
             <FormControl>
               <RadioGroup
                 aria-labelledby='demo-radio-buttons-group-label'
-                defaultValue='female'
                 name='radio-buttons-group'
```
fengelniederhammer: Do we need the `aria-labelledby` and the `name`? Or could we delete those, too?

src/components/MutationTableTabContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+          name: variant.name.length > 0 ? variant.name : formatVariantDisplayName(variant.query),
+          variantCount: variant.variantCount,
+          aaMutations: variant.aaMutations,
+          nucMutations: variant.nucMutations,
+        };
+      });
+    }
+  }, [mutations]);
+
+  const baselineTableData = useMemo(() => {
```
fengelniederhammer: Why do we actually need a `useMemo` here? Couldn't we simply write 
```
const baselineTableData = mode === 'CompareToBaseline' && baselineMutations.data && mutations.data 
  ? baselineMutations.data 
  : undefined
``` 
?

fengelniederhammer: Why do we need that check at all? Isn't is essentially the same check as line 132 below?

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,271 @@
...
+  baseline,
+  dateRangeSelector,
+  baselineVariant,
+}: NameCellProps) => {
+  const urlParams = new URLSearchParams();
+  addVariantSelectorToUrlSearchParams(query, urlParams);
+
+  const placeString = encodeLocationSelectorToSingleString(locationSelector);
+  return (
+    <>
```
fengelniederhammer: We don't need that fragment

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,271 @@
...
+const columnNames: string[] = mutationColumnsProperties.map(item => item.label);
+
+export const SingleMutationsTable = ({
+  mutations,
+  locationSelector,
+  mutationType,
+  dateRangeSelector,
+}: SingleMutationsTableProps) => {
+  return (
+    <>
```
fengelniederhammer: We don't need that fragment

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,271 @@
...
+export const BaselineMutationsTable = ({
+  baselineData,
+  mutations,
+  mutationType,
+  locationSelector,
+  baselineVariant,
+  dateRangeSelector,
+}: BaselineMutationsTableProps) => {
+  const headers: string[] = ['Additional', 'Missing'];
+  console.log('baseline data', baselineData);
```
fengelniederhammer: Please delete that log statement before merge.

src/components/MutationsTable.tsx
```diff
@@ -0,0 +1,271 @@
...
+              {columnNames.map((name, i) => (
+                <td key={`${name}-${i}`}>{name}</td>
+              ))}
+              {columnNames.map((name, i) => (
+                <td key={`${name}/${i}`}>{name}</td>
+              ))}
+            </tr>
+          </thead>
+          <tbody>
+            {mutations &&
```
fengelniederhammer: Same here: Please require in the Props that mutations is not undefined and remove this check here.

src/components/VariantMutations.tsx
fengelniederhammer: I don't know how to fix this or whether it is possible to fix this at all with simple means. I guess that the issue is that the page loads tons of data and builds up incrementally, thus it scroll to the "correct" place, but after loading the next diagram, the actually correct place is further down.
How much of a hard requirement is this? Could we simply leave out this feature? Or implement something else instead (only showing the targetted component)? I think we should at least not leave a non-working feature here.

src/components/MutationsTable.tsx
fengelniederhammer: Ok, makes sense. Please add a description to the table to clarify this. @chaoran-chen said he will provide a text at the beginning of next week.

src/data/api-lapis.ts
```diff
@@ -27,15 +27,12 @@ import { LapisSelector } from './LapisSelector';
...
 import { addQcSelectorToUrlSearchParams } from './QcSelector';
 import { HostCountSampleEntry } from './sample/HostCountSampleEntry';
-import { sequenceDataSource } from '../helpers/sequence-data-source';
 import { InsertionCountEntry } from './InsertionCountEntry';
 import { NextcladeDatasetInfo } from './NextcladeDatasetInfo';
 
 const HOST = process.env.REACT_APP_LAPIS_HOST;
 const ACCESS_KEY = process.env.REACT_APP_LAPIS_ACCESS_KEY;
 
-export const HUMAN = sequenceDataSource === 'gisaid' ? 'Human' : 'Homo sapiens';
```
fengelniederhammer, line 37: I moved this to resolve a circular module dependency

src/components/TableTabContent.tsx
```diff
@@ -0,0 +1,263 @@
+import { DataGrid, GridColDef, GridComparatorFn, GridRenderCellParams } from '@mui/x-data-grid';
```
fengelniederhammer, line 1: This was extracted from the `CollectionSingleViewPage`

src/components/SequencesOverTimeTabContent.tsx
```diff
@@ -0,0 +1,103 @@
+import { useRef } from 'react';
```
fengelniederhammer, line 1: This was extracted from the `CollectionSingleViewPage`

src/components/VariantMutations.tsx
fengelniederhammer: I set a timeout after which the scroll effect happens.

fengelniederhammer:
@chaoran-chen Could you please provide the explanation text for the baseline comparison table? Otherwise this PR is ready from our side and waits for a review :)

Sorry that this PR got that messy, I had to rebase it and squashed all commits. What this PR includes at this point:
* some test setup: Mock the API calls that are issued at static "load the file" time. This should be refactored at some point in the future.
* resolving circular dependencies of modules that somehow broke the page (I don't know why it worked before - there are still some circular dependencies)
* The new mutation tables - structured in hooks that compute the data and components that only display a table, including tests.
* Extraction of some components to make the `CollectionSingleViewPage` smaller.

fengelniederhammer (approved):

chaoran-chen:
@fengelniederhammer, I added the text to the mutation table.

chaoran-chen (approved):

PR #661, opened 2023-01-03
Pass 5k instead of 1k seqs to Usher, use 5k instead of 1k sequences in context

corneliusroemer:
Usher is capable of making a tree with up to 5000 sequences. Currently, we tell Usher to make a tree with 1000, which is unnecessarily restrictive and limits the usefulness without good reason - I always use Usher with 5k context sequences for my usage and never encounter any problems.

Maybe @AngieHinrichs can advise if there are any issues with using 5k instead of 1k sequences. 

Actually, it turns out I misunderstood. This PR now shares 5k sequences to Usher which Usher amazingly has no problem dealing with - I managed to get the entire BA.2.3.20* tree with ~5k sequences. Hats off!

What I actually wanted to achieve is to get more _context_ sequences, but this is cool as well. @AngieHinrichs, how can we include more context samples? The number box here of samples to show per tree? It's actually nice to be able to get trees bigger than that (in my case 6.5k here)
image

image

### DIFF
--- a/src/services/external-integrations/UsherIntegration.ts
+++ b/src/services/external-integrations/UsherIntegration.ts
@@ -7,7 +7,7 @@ import { LapisSelector } from '../../data/LapisSelector';
 const usherUrl =
   'https://genome.ucsc.edu/cgi-bin/hgPhyloPlace?db=wuhCor1&phyloPlaceTree=hgPhyloPlaceData/wuhCor1' +
   '/public.plusGisaid.latest.masked.pb&subtreeSize=5000&remoteFile=';
-const defaultOrderAndLimit: OrderAndLimitConfig = { orderBy: 'random', limit: 5000 };
+const defaultOrderAndLimit: OrderAndLimitConfig = { orderBy: 'random', limit: 1000 };
 
 export class UsherIntegration implements Integration {
   name = 'UShER';
--- a/src/services/external-integrations/UsherIntegration.ts
+++ b/src/services/external-integrations/UsherIntegration.ts
@@ -7,7 +7,7 @@ import { LapisSelector } from '../../data/LapisSelector';
 const usherUrl =
   'https://genome.ucsc.edu/cgi-bin/hgPhyloPlace?db=wuhCor1&phyloPlaceTree=hgPhyloPlaceData/wuhCor1' +
   '/public.plusGisaid.latest.masked.pb&subtreeSize=1000&remoteFile=';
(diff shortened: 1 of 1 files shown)


src/services/external-integrations/UsherIntegration.ts
```diff
@@ -6,8 +6,8 @@ import { LapisSelector } from '../../data/LapisSelector';
 
 const usherUrl =
   'https://genome.ucsc.edu/cgi-bin/hgPhyloPlace?db=wuhCor1&phyloPlaceTree=hgPhyloPlaceData/wuhCor1' +
-  '/public.plusGisaid.latest.masked.pb&subtreeSize=1000&remoteFile=';
-const defaultOrderAndLimit: OrderAndLimitConfig = { orderBy: 'random', limit: 1000 };
+  '/public.plusGisaid.latest.masked.pb&subtreeSize=5000&remoteFile=';
+const defaultOrderAndLimit: OrderAndLimitConfig = { orderBy: 'random', limit: 5000 };
```
corneliusroemer: Actually it's probably best to leave this at 1k, and just up the subtreeSize
It's a bit silly to peruse 5k sequences, I've never wanted to look at that many sequences at once.

What matters is having a lot of context, the more context the better.

corneliusroemer:
```suggestion
const defaultOrderAndLimit: OrderAndLimitConfig = { orderBy: 'random', limit: 1000 };
```

corneliusroemer:
Tested a bit with common use cases and it works well with current settings.

AngieHinrichs:
> What I actually wanted to achieve is to get more context sequences, but this is cool as well. @AngieHinrichs, how can we include more context samples? The number box here of samples to show per tree? It's actually nice to be able to get trees bigger than that (in my case 6.5k here)

The "Number of samples per subtree" input is the same as the `subtreeSize` request parameter -- that number is passed to `matUtils extract` as the `-x` (`--usher-minimum-subtrees-size`) parameter.  That number of samples are added as context for each subtree, so for example in your screenshot, 4919 uploaded sample names were placed in the same subtree and then 2000 context samples were added for a total of 6919.  The upper limit of 5000 is pretty arbitrary and is based on my assumption of the upper limit that the auspice display can handle comfortably.

Actually, I think the limit of 5000 is enforced only by the web page JS, not the back end.  So... you could try passing a larger subtreeSize and see how well it displays.  🙂  (But also think of how it will display for users with older slower PCs)

chaoran-chen (approved):
Thanks @corneliusroemer and @AngieHinrichs!

PR #665, opened 2023-01-09
change mutation sorting function

### LINKED ISSUE #587: Sort AA mutations by gene position
Currently, in the variant mutation table if "sort by position" is chosen, the entries are first sorted alphabetically by the gene name and then by the position within each gene. It would however be better if the genes are sorted by their positions in the genomic sequence.

anastasia-escher:
Closes #587

### DIFF
--- a/src/components/VariantMutations.tsx
+++ b/src/components/VariantMutations.tsx
@@ -204,6 +204,7 @@ export const VariantMutations = ({ selector }: Props) => {
         }),
         exportManager.register('Download nucleotide consensus sequence (ignoring deletions)', async () => {
           const reference = (await ReferenceGenomeService.data).nucSeq;
+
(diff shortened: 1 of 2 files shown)


chaoran-chen (changes requested):
Hi @anastasia-escher. [This preview page](https://cov-spectrum-develop-git-mutationsort-cov-spectrum.vercel.app/explore/Switzerland/AllSamples/Past6M/variants?variantQuery=nextcladePangoLineage%3Aba.5*+%26+%21nextcladePangoLineage%3ABQ.1*&) looks like this to me:

image

Isn't it very wrong? If you look at [refData](https://github.com/GenSpectrum/cov-spectrum-website/blob/develop/src/data/refData.json#L3) file, you can see that ORF1a is at the beginning of the genome, followed by ORF1b and S.

PR #672, opened 2023-01-17
get rid of warnings in console while executing the tests

fengelniederhammer:
The tests complained that the ExportManager was not set via the context provider.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -90,6 +90,7 @@
         "@types/react-router-hash-link": "^2.4.2",
         "all-contributors-cli": "^6.20.0",
         "autoprefixer": "^10.4.13",
+        "cross-env": "^7.0.3",
         "eslint": "^8.21.0",
         "eslint-plugin-unused-imports": "^2.0.0",
         "jest-canvas-mock": "^2.3.1",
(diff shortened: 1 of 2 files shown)


chaoran-chen:
Oh, sorry that it's not documented: these warnings shouldn't occur if you set the environment variable `REACT_APP_IS_TESTING=true`

fengelniederhammer:
Why do we need an environment variable to control that? Doesn't it make more sense to test the components in a more realistic scenario, i.e. using the context providers? Then the tests could actually control the ExportManagers.

chaoran-chen:
Good question. The disadvantage of this solution is that the charts will always be tested together with the ExportManager although they can be used alone.

I wonder whether a good solution might be to just remove the warnings ([code](https://github.com/GenSpectrum/cov-spectrum-website/blob/develop/src/components/CombinedExport/ExportManager.ts#L21)). I don't see big harm in calling `ExportManager.register()` even if there is no context provider - it's then just a no-op.

Ideally, I guess that the chart components should all be refactored to better separate the calculation logic, the view, and the DownloadWrapper (which calls the ExportManager). But that would take some effort

chaoran-chen (approved):
Looks good, thanks!

PR #678, opened 2023-01-25
Show what a recombinant alias stand for #585

### LINKED ISSUE #585: Show what a recombinant alias stand for
For Pango lineages, we display the resolved alias if it is not a recombinant:

We should add an explanation for recombinants (Pango lineages that start with an "X") as well. A few suggestions:

- For XAA: Recombinant of BA.1* and BA.2*
- For XAA.1.5: Child of recombinant of BA.1* and BA.2*
- For XAC (which has `["BA.2*","BA.1*","BA.2*"]` as parents), I suggest mentioning each parent only once. I.e., "Recombinant of BA.2* and BA.1*"

The recombinant data can be retrieved from the `/resource/pango-lineage-recombinant` endpoint of the CoV-Spectrum server ([docs](https://github.com/GenSpectrum/cov-spectrum-server/blob/develop/docs/api.md#pango-lineage-recombinants)).

JonasKellerer:
Resolves #585

Further refactoring is possible for PangoLineageAliasResolverService but not attempted because it is nested into custom useQuery and thus too much work for now. Actually the PangoLineageAliasResolverService should be replaced by usePangoLineageFullName and usePangoLineageWithAlias.

### DIFF
--- a/src/components/VariantHeader.tsx
+++ b/src/components/VariantHeader.tsx
@@ -1,6 +1,8 @@
-import React, { useEffect, useState } from 'react';
-import { PangoLineageAliasResolverService } from '../services/PangoLineageAliasResolverService';
-import { formatVariantDisplayName, VariantSelector } from '../data/VariantSelector';
+import React, { useMemo } from 'react';
+import { formatVariantDisplayName, getPangoLineage, VariantSelector } from '../data/VariantSelector';
+import { fetchPangoLineageRecombinant } from '../data/api';
+import { usePangoLineageFullName } from '../services/pangoLineageAlias';
+import { useQuery } from '../helpers/query-hook';
 
 export interface Props {
   variant: VariantSelector;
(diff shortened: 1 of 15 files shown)


src/components/VariantHeader.tsx
```diff
@@ -1,38 +1,61 @@
...
-      : variant.nextcladePangoLineage
-      ? variant.nextcladePangoLineage
-      : '';
-    PangoLineageAliasResolverService.findFullName(pangoLineage).then(name => {
-      if (isSubscribed) {
-        setResolvedFullName(name);
+  if (lineageAliasText !== undefined) {
+    return lineageAliasText;
+  }
+  return recombinationText;
```
chaoran-chen: this could be simplified to `return lineageAliasText ?? recombinationText`

chaoran-chen (changes requested):
Thanks! It looks quite good. I just have one tiny request: For a Pango lineage query ending with a `*`, the same text should be displayed as for the Pango lineage query without it.

image
image

And I've one question: Why did you decide to not continue using the `PangoLineageAliasResolverService`? It was ensuring that the aliases are only loaded once.

fengelniederhammer:
> And I've one question: Why did you decide to not continue using the `PangoLineageAliasResolverService`? It was ensuring that the aliases are only loaded once.

The problem with it was that it was almost impossible to test because of the static initalization. We fiddled around with it for 2 hours until we decided to rewrite it in a way that it can easily be tested. Actually we would like to migrate all usages of the `PangoLineageAliasResolverService` to our custom hook, with `useQuery` ensuring that it's only loaded once.

chaoran-chen:
Ok, I see! Now, I also understand better why you wanted to introduce a cache-capable query library. For now, I think that it's fine to make a few additional requests. We can then talk in our next meeting about ways to improve :)

JonasKellerer:
What does the * actually mean? Is it a matchAll function?

chaoran-chen:
`*` means that all sub-lineages are included. `B.1.1` selects sequences that are labeled with exactly `B.1.1`. `B.1.1*` includes`B.1.1` and all children of `B.1.1` such as `B.1.1.1`, `B.1.1.529`, `BA.1`, and many more.

(Check out https://cov-spectrum.org/about#faq-search-variants (section "How can I search a variant?"))

chaoran-chen (approved):

PR #679, opened 2023-01-26
Buttons in the advanced filters panel are too big #677

### LINKED ISSUE #677: Buttons in the advanced filters panel are too big
image

Not intended but somehow an interesting style :)

JonasKellerer:
Fix #677


chaoran-chen (changes requested):
Thanks, @JonasKellerer! This looks good to me. Just one small request: I think that we should keep the abbreviation "qc". It is quite common and also used in LAPIS and by the underlying tool Nextclade.

Maybe we could create a glossary in the docs to explain it?

chaoran-chen (approved):

PR #684, opened 2023-02-02
change date language to english

### LINKED ISSUE #577: Week date in update date is in German
Why is the week day written in German? It should always be in English.

image

anastasia-escher:
resolves #577

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -50,6 +50,7 @@ const Footer = styled.footer`
 `;
 
 export const App = () => {
+  dayjs.locale('en');
   const [hideHeaderAndFooter, setHideHeaderAndFooter] = useState(false);
   const { width, ref } = useResizeDetector<HTMLDivElement>();
   const isSmallScreen = width !== undefined && width < 768;


JonasKellerer:
The fix works for me. 

The `DatePicker` has their first week day still at sunday (as expected for en). If we would want to change this, there is the option to add `adapterLocale={theLocaleFromBrowser}' to the `LocalizationProvider` in DateRangePicker.tsx .

fengelniederhammer:
This contradicts the following: https://github.com/GenSpectrum/cov-spectrum-website/blob/f89e46534e9d68e6bf4b5e96f1078997116a26f6/src/helpers/date-cache.ts#L10

fengelniederhammer (changes requested):
This contradicts this line:
https://github.com/GenSpectrum/cov-spectrum-website/blob/f89e46534e9d68e6bf4b5e96f1078997116a26f6/src/helpers/date-cache.ts#L10

#726 provides an alternative solution.

PR #709, opened 2023-02-09
fix: When changing time range for country with space in name, page resets

### LINKED ISSUE #707: Bug: When changing time range for country with space in name, page resets
When selecting a different date range, the current query is lost:

I could reproduce on Firefox. But somehow it may be country specific: Switzerland works. Maybe something about spaces again?

Seems to be regression of https://github.com/GenSpectrum/cov-spectrum-website/issues/704 and https://github.com/GenSpectrum/cov-spectrum-website/issues/694

fengelniederhammer:
resolves #707 

I hope we caught all countries in URLs this time. We might have to look out for more unexpected characters in URLs in other places though.


chaoran-chen (approved):
Thank you very much! I'll merge it right away because it's an urgent fix.

PR #713, opened 2023-02-13
When not using advanced search, first character entered for each term gets ignored.

### LINKED ISSUE #696: When not using advanced search, first character entered for each term gets ignored.
ie, if I want to search "s:486p, s:444t" I will have to type "ss:486p" enter, "ss:444t". Visually everything displays correctly, but the first button press doesn't make it in at all. 

I'm using google chrome, haven't tested on other browsers.

JonasKellerer:
Resolves: #696

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,6 +57,7 @@
         "react-dnd": "^16.0.1",
         "react-dnd-multi-backend": "^8.0.0",
         "react-dnd-preview": "^8.0.0",
+        "react-dnd-test-backend": "^16.0.1",
         "react-dnd-touch-backend": "^16.0.1",
         "react-dom": "^18.2.0",
         "react-icons": "^4.7.1",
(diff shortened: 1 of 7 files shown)


chaoran-chen (approved):
Thank you! It works as far as I can tell!

Would you happen to know what triggered the bug? Why did it not occur in the past?

chaoran-chen:
I merged it as it is an urgent bug fix.

corneliusroemer:
Amazing, this works in my testing (tried out a number of edge cases as well!)

fengelniederhammer:
> Would you happen to know what triggered the bug? Why did it not occur in the past?

We don't know for sure what caused the bug. We're pretty sure that it was introduced by the React 18 upgrade (on https://cov-spectrum-develop-git-fixwarnings-cov-spectrum.vercel.app/explore/Germany/AllSamples/Past6M it still works). 
We deleted that "if menu closed, then delete input" statement, because we found it quite weird and unintuitive anyway. That definitely caused the bug. So actually the question is, why did it work before at all? ;)

FedeGueli:
great work thank you guys. we had already got used with that ;)

PR #714, opened 2023-02-13
fix: variant search field sometimes flickering back and forth

### LINKED ISSUE #711: Adding a new term to an advanced search can cause the text to flicker back and forth..?
I'm definitely missing some details on this one. It should be highly reproducible yet inconsistent though. 

Version A:
When I'm editing collections such as 37, when I go to add a new term to my advances queries (ie, !s:147e), I can enter a few characters without issue, but then a few after that will have an issue where they flicker back and forth. Further characters I add to it after this also flicker back and forth. (meaning, they appear and then disappear rapidly). This issue resolves itself unpredictably after I add or delete some text. The issue does not resolve itself on its own. The issue can come back without me exiting the field. 
This can also happen immediately after switching a query to be regular to advanced, and trying to add a new term. 

Version B: 
I can cause the issue just by doing an advanced search on the main site, outside of a collection. Add in a term, search, add in another term, search, add in another term, search, etc. I strongly suspect this comes down to speed/if you are trying to add terms while the system is still processing something.. from what I can tell, if you click search and then go and add something (ie, &s:452m) quickly, the issue will happen. Additionally, I have gotten the issue to happen a similar way but by clicking the "advanced search" button to toggle between it. This causes the button to flicker back and forth. 

I believe this issue may have started at the same time as #696, but the issue definitely did not exist on Feb 2.

fengelniederhammer:
resolves #711

We are not completely sure whether this fixes the issue, but this [`useEffect`](https://github.com/GenSpectrum/cov-spectrum-website/blob/d71e817700548b1f22fe3745d1e2ed2a21e96208/src/components/VariantSearchField.tsx#L361-L370) looked very suspicious. After removing it, we could not reproduce the issue anymore.

It's not so obvious from the diff: We mainly deleted the mentioned `useEffect` and refactored the code to perform the `onVariantSelect` calls directly where the state is set.

### DIFF
--- a/src/components/VariantSearchField.tsx
+++ b/src/components/VariantSearchField.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import { useState } from 'react';
 import AsyncSelect from 'react-select/async';
 import { components, CSSObjectWithLabel, InputActionMeta, StylesConfig } from 'react-select';
 import { isValidAAMutation, isValidABNotation, isValidNspNotation } from '../helpers/aa-mutation';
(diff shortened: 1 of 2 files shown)


chaoran-chen (approved):
Thanks! Looks good to me!

PR #721, opened 2023-02-14
fix: make tooltip in world map in Omicron story visible

fengelniederhammer:
dark letters on black background are not readable


chaoran-chen:
Almost forgot that it existed. Let's do that instead: #722

PR #723, opened 2023-02-14
build: use depcheck to check for unused/missing dependencies

fengelniederhammer:
Please provide feedback whether you like it.

Especially the tool complains about:
```
Unused devDependencies
* all-contributors-cli
* autoprefixer
```
Are those still in use?


chaoran-chen:
I think it's a great idea!

I use `all-contributors-cli` to add people to the "contributors" section in README.md. I'm not sure about `autoprefixer`. I think that it came along with tailwindcss.

JonasKellerer:
depcheck wanted to remove `@types/jest`j so I did. It seems that it was indeed not needed on my machine for the tests. Could you check if this is also so case for you? 

- [x] the assert package is still in, so the build fails, but this is addressed in #724 .

chaoran-chen (approved):
Looks good to me!

PR #724, opened 2023-02-15
"assert" library is outdated and not working properly

### LINKED ISSUE #695: "assert" library is outdated and not working properly
We have a number of `assert()` statements using the ["assert" package](https://www.npmjs.com/package/assert). That package, however, isn't maintained anymore and not working properly. When an assertion fails, it throws an "assertion_error.js:392 Uncaught ReferenceError: process is not defined" error. We should get rid of the package.

JonasKellerer:
Resolves: #695


chaoran-chen (approved):
Looks good!

PR #725, opened 2023-02-15
Remove Omicron story

### LINKED ISSUE #722: Remove Omicron story
The [Omicron story](https://cov-spectrum.org/stories/omicron) page can be removed. It's not useful anymore (and not sure if it has ever been).

JonasKellerer:
Resolves: #722


fengelniederhammer:
The `VariantInternationalComparisonMap` is now unused (including the world map). @chaoran-chen Should we delete it? Or is there any chance that it will be used somewhere else in the future?

chaoran-chen:
Eventually, it would be nice to add more maps including a world map to the "normal" variant pages but I don't think that we will do it very soon, so I think that we can remove the component.

src/maps/RegionMap.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
```
fengelniederhammer: I think we can delete the whole file, can't we? Including the `world.json` next to it, as well as the dependency on `react-simple-maps`?

JonasKellerer: Yes. It is no longer needed in other files.

fengelniederhammer:
Looks good to me :+1:

chaoran-chen (approved):
👍

PR #726, opened 2023-02-15
fix: don't display the weekday in footer in German

### LINKED ISSUE #577: Week date in update date is in German

fengelniederhammer:
resolves #577

This is an alternative solution to #684.


chaoran-chen (approved):
Yes, let's do it this way. Thanks!

PR #736, opened 2023-02-22
fix: remove dependency on react-router, use react-router-dom instead.

fengelniederhammer:
See https://github.com/remix-run/react-router/blob/f0b886bce7dab2e2f37f53fe58cd6c7ab0660e25/packages/react-router/README.md

The problem appeared in #729


chaoran-chen (approved):
Thanks for the fix!

src/__tests__/App.test.tsx
```diff
@@ -0,0 +1,33 @@
+import { act, render, screen } from '@testing-library/react';
+import { App } from '../App';
+import { MemoryRouter } from 'react-router-dom';
+import ResizeObserver from 'resize-observer-polyfill';
+import { useResizeDetector } from 'react-resize-detector';
+
+jest.mock('../data/api');
+jest.mock('../data/api-lapis');
```
fengelniederhammer, line 8: This test needs the api mocks from the mutation overview branch, so I rebased it on top of it. We'll have to merge this after the base branch is merged.

PR #776, opened 2023-04-12
Fix mobile view

### LINKED ISSUE #772: Strange mobile landing page view
I just noticed that on my phone if I open https://cov-spectrum.org, the view is correct (screenshot 1) and if I open https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M, it is wrong (screenshot 2). Thereby, the first link redirects to the second.

Maybe I broke something in fa6d764ef87f7de6883a7d5d99d95c97d92a842a?

**Screenshot 1**

**Screenshot 2**

fengelniederhammer:
resolves #772 

Do the "base location is set" check after setting the ref for the screen width. Otherwise the ref remains uninitialized and the screen width is recognized too late.

Reviewing the individual commits separately should be easier to follow along.


chaoran-chen (approved):
Perfect! Thanks for cleaning up the code!

PR #779, opened 2023-04-17
fix: chat history browsing with arrow up and down

### LINKED ISSUE #777: History browsing in chat in Firefox is misbehaving
The chat message input bar does not work very well in Firefox. One issue:

1. open `/chat` (include access key)
2. Click "yes" or "no"
3. Enter an arbitrary question or click on the random box
4. Wait for the answer
5. While the cursor is in the input, press the buttons: "Arrow Up", "Arrow Down", "Arrow Down", "Arrow Up"

**Expected behavior/behavior in Chrome:**

The previous is in the input box.

**Behavior in Firefox:**

The input box is empty.

This is probably somehow related to https://github.com/chatscope/chat-ui-kit-react/issues/91#issuecomment-1504848904 and [my miserable implementation of the history browsing](https://github.com/GenSpectrum/cov-spectrum-website/blob/ebcb2c19a9b08cc5e93e8823beed9274eb1009b5/src/components/chat/CustomMessageInput.tsx#L73).

**Optional task:** Check other browsers

fengelniederhammer:
resolves #777

Firefox behaves weirdly if the "empty" contenteditable div contains an empty string. Firefox by default puts at <br> there, even if the div is "empty".

### DIFF
--- a/src/components/chat/CustomMessageInput.tsx
+++ b/src/components/chat/CustomMessageInput.tsx
@@ -14,16 +14,18 @@ export type CustomMessageInputProps = {
 type MessageInputTriplet = {
   innerHtml: string;
   textContent: string;
-  innerText: string;
+  innerText: string | undefined;
+};
+
+const EMPTY_CHAT_INPUT = {
+  innerHtml: '<br>',
+  textContent: '<br>',
+  innerText: '',
 };
 
 export const CustomMessageInput = ({ disabled, maxLength, onMessageSend }: CustomMessageInputProps) => {
   const [messageInputRef, setMessageInputFocus] = useFocus();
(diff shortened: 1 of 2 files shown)


chaoran-chen (changes requested):
Thanks, this fixes the bug! However, the placeholder is unfortunately also gone. Would it be possible to have both?

Before:

image

Now:

image

chaoran-chen:
I'll already merge it because having a working history browsing is more important than the placeholder.

fengelniederhammer:
Yes, sorry, I didn't find a way to make it work. Either the field contains `<br>` and the navigation works in Firefox, but there is no placeholder, or the field is truly empty and shows the placeholder, but then the navigation in Firefox is broken.

So actually I think that Firefox is broken here.

PR #782, opened 2023-04-18
Chat: Improve results tables

### LINKED ISSUE #778: Chat: improve result tables
That's how the tables currently look like:

Improvements:

- [x] Make the tables sortable (use material UI's data table again?)
- [x] Proportions should be formatted as percentages with two decimal places
- [ ] Check the UI/UX on mobile and improve if needed
- [x] Add a subtle download icon to download the data as a CSV

JonasKellerer:
Resolves #778

### DIFF
--- a/src/components/chat/IncomingResponseMessage.tsx
+++ b/src/components/chat/IncomingResponseMessage.tsx
@@ -40,7 +40,7 @@ type ChatDataTableProps = {
 
 const ChatDataTable = ({ data }: ChatDataTableProps) => {
   const [paginationModel, setPaginationModel] = React.useState({
-    pageSize: 5,
+    pageSize: 10,
     page: 0,
   });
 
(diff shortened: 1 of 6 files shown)


JonasKellerer:
'Check the UI/UX on mobile and improve if needed' is still open. I checked it through the browser tools, where it seems okay. I think not needing to scroll due to the pages helps. Feedback is very welcome.

src/components/chat/IncomingResponseMessage.tsx
```diff
@@ -27,30 +27,87 @@ export const IncomingResponseMessage = ({
...
+        columns={headers}
+        rows={rows}
+        autoHeight={true}
+        paginationModel={paginationModel}
+        pageSizeOptions={[5, 10, 20]}
+        onPaginationModelChange={model => setPaginationModel(model)}
+        sx={{
+          backgroundColor: 'background.paper',
+        }}
+        slots={{ toolbar: CustomToolbar }}
```
chaoran-chen: I think that it would be better to save some space and make the rows as narrow as possible

chaoran-chen (changes requested):
Thank you very much! Only one small request (see below), otherwise it's perfect!

chaoran-chen:
Oh, and I don't think that we need the "print" option. Let's only offer the CSV download for now.

chaoran-chen (approved):

PR #784, opened 2023-04-21
Nucleotide Diversity

philipschulz:
- Nucleotide Diversity chart component for Focus page
- Nucleotide Diversity multi chart component for CompareEquals page


src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+import { useExploreUrl } from '../helpers/explore-url';
+import { pprettyColors } from '../helpers/colors';
+import { getTicks } from '../helpers/ticks';
+import { formatDate } from '../widgets/VariantTimeDistributionLineChartInner';
+
+type Props = {
+    selectors: LapisSelector[];
+  };
+
+type Variant = {
```
fengelniederhammer: Do we need this type? According to my IDE it is unused.

src/components/NucleotideEntropyMultiChart.tsx
fengelniederhammer, line 1: eslint (`npx eslint .`) complains about a couple of lines here and in `NucleotideEntropy.tsx`. Could you please fix those?

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+    const data = useData(selectors, selectedGene, variants, sequenceType, deletions);
+
+    const controls = (
+        <div className='mb-4'>
+          {/* AA vs. nucs */}
+          <div className='mb-2 flex'>
+            <PipeDividedOptionsButtons
+              options={[
+                { label: 'NUC', value: 'nuc' },
+                { label: 'AA', value: 'aa' },
```
fengelniederhammer: The surrounding diagrams use the full names.
```suggestion
                { label: 'Nucleotides', value: 'nuc' },
                { label: 'Amino acids', value: 'aa' },
```

src/components/NucleotideEntropyMultiChart.tsx
fengelniederhammer, line 1: Please use prettier to format the files. If you use IntelliJ, there is a nice plugin :)

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+  genes.push({ name: 'All', startPosition: 0, endPosition: 29903, aaSeq: ''});
+
+function nonNullValues(obj: any) {
+  return Object.entries(obj).filter(([key, value]) => value !== undefined).map(([key, value]) => value).toString();
+}
+
+export const NucleotideEntropyMultiChart = ({ selectors }: Props) => {
+    const [sequenceType, setSequenceType] = useState<SequenceType>('nuc');
+    const [gene, setGene] = useState<string>('All');
+    const [deletions, setDeletions] = useState<boolean>(false);
```
fengelniederhammer:
```suggestion
    const [includeDeletions, setIncludeDeletions] = useState<boolean>(false);
```
That would make it more obvious that the variable is a boolean.

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+              </Form.Control>
+            </div>
+          }
+        </div>
+    );
+
+
+
+    let plotArea;
+    if (variants![0].length == 0) {
```
fengelniederhammer: That `!` looks supsicious to me. This basically disables typesafety. Why is the type nullable if you seem to be sure that it is not null here?

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+            aaSeq: g.aaSeq,
+            color: ''
+          };
+    })
+
+    const data = useData(selectors, selectedGene, variants, sequenceType, deletions);
+
+    const controls = (
+        <div className='mb-4'>
+          {/* AA vs. nucs */}
```
fengelniederhammer: Do we need those comments here? I'd propose to
* either simply delete them
* or extract the code that the comment refers to into a separate component and give that component a name that replaces the comment.

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+                    checked={deletions}
+                    onChange={handleDeletions}
+                    inputProps={{ 'aria-label': 'controlled' }}
+                  />
+                }
+                label='Include deletions'
+              />
+            </FormGroup>
+          {/* Genes */}
+          {
```
fengelniederhammer: I think we don't need this pair of brackets, do we?

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+            <div className=' w-72 flex mb-2'>
+              <div className='mr-2'>Gene:</div>
+              <Form.Control
+                as='select'
+                value={gene}
+                onChange={ev => setGene(ev.target.value)}
```
fengelniederhammer: What is `ev`? Could you please give it a proper name? :)

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+    selectedGene: GeneOption[],
+    variants: any,
+    sequenceType: SequenceType,
+    deletions: boolean
+): any | undefined => {
+  //fetch the proportions per position in weekly segments
+  const weeklyVariantMutationProportionQuery = useQuery(
+    async signal => {
+      //calculate weeks
+      const dayArray: UnifiedDay[] = [selectors[0].dateRange?.getDateRange().dateFrom!, selectors[0].dateRange?.getDateRange().dateTo!];
```
fengelniederhammer: Explicit type declarations are usually not necessary. Typescript is smart enough to figure out the type by itself.
```suggestion
      const days = [selectors[0].dateRange?.getDateRange().dateFrom!, selectors[0].dateRange?.getDateRange().dateTo!];
```

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+      const dayRange = globalDateCache.rangeFromDays(dayArray)!;
+      const weeks = globalDateCache.weeksFromRange({ min: dayRange.min.isoWeek, max: dayRange.max.isoWeek });
+      const weekDateRanges = new Array<DateRange>();
+      for (let i = 0; i < weeks.length; i++) {
+        let dateRange: DateRange = { dateFrom: weeks[i].firstDay, dateTo: weeks[i].firstDay };
+        weekDateRanges[i] = dateRange;
+      }
```
fengelniederhammer: Using `Array.map` is usually more idiomatic are easier to read:
```suggestion
      const weekDateRanges: DateRange[] = globalDateCache
        .weeksFromRange({
          min: dayRange.min.isoWeek,
          max: dayRange.max.isoWeek,
        })
        .map(week => ({
          dateFrom: week.firstDay,
          dateTo: week.firstDay,
        }));
```

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+              };
+            })
+          )
+        )
+      }
+    },
+    [selectors, sequenceType]
+  );
+    
+  const data = useMemo(() => {
```
fengelniederhammer: This could be inlined:
```suggestion
  return useMemo(() => {
```

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+      const weekSelectors = new Array<LapisSelector>(); 
+      selectors.forEach(selector => {
+          for (let w of weekDateRanges) { 
+            weekSelectors.push({
+            ...selector,
+            dateRange: new FixedDateRangeSelector(w),
+          })
+        };
+      })
```
fengelniederhammer: Same here: (Leveraging ChatGPT :) )
```suggestion
        const weekSelectors = selectors.flatMap(selector =>
            weekDateRanges.map(w => ({
                ...selector,
                dateRange: new FixedDateRangeSelector(w),
            })),
        );
```

fengelniederhammer: This is the explanation that ChatGPT gave:
This version leverages the flatMap function to flatten the resulting nested arrays into a single-level array while applying the provided mapping function. In this case, for each selector element, we create a new array of LapisSelector objects with the updated dateRange property, and then flatten the resulting arrays into one continuous array.

src/components/NucleotideEntropyMultiChart.tsx
fengelniederhammer: Could you please double check all the `!` in your code? This potentially results in errors.

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+      let weekLength = weekDateRanges.length;
+
+      return {
+        sequenceType: sequenceType,
+        weekNumber: weekLength,
+        result: await Promise.all(
+          weekSelectors.map((w, i) =>
```
fengelniederhammer: Let's also use proper variable names here, please: 
```suggestion
          weekSelectors.map((weekSelector, i) =>
```

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+            MutationProportionData.fromApi(w, sequenceType, signal).then(data => {
+              const proportions: MutationProportionEntry[] = data.payload.map(m => m);
```
fengelniederhammer: Why do we need this line? It essentially does nothing.

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+const useData = (
+    selectors: LapisSelector[],
+    selectedGene: GeneOption[],
+    variants: any,
+    sequenceType: SequenceType,
+    deletions: boolean
+): any | undefined => {
+  //fetch the proportions per position in weekly segments
+  const weeklyVariantMutationProportionQuery = useQuery(
```
fengelniederhammer: Let's also get rid of the comments here: :) 
```suggestion
  const weeklyVariantMutationProportions = useProportionsPerPositionInWeeklySegments(selectors, sequenceType);

```

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+    },
+    [selectors, sequenceType]
+  );
+    
+  const data = useMemo(() => {
+    if (!weeklyVariantMutationProportionQuery.data) {
+      return undefined;
+    }
+    let sequenceType = weeklyVariantMutationProportionQuery.data.sequenceType;
+    if (!sequenceType) {
```
fengelniederhammer: Can `sequenceType` ever be falsy?

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+            ...selector,
+            dateRange: new FixedDateRangeSelector(w),
+          })
+        };
+      })
+
+      let weekLength = weekDateRanges.length;
+
+      return {
+        sequenceType: sequenceType,
```
fengelniederhammer: Why do you return `sequenceType` here? This is the same as the input.

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+      return undefined;
+    }
+    if (!selectedGene) {
+      return undefined;
+    }
+    if (!variants[0]){
+      return undefined;
+    }
+    if (variants.length*weekNumber != weeklyVariantMutationProportionQuery.data.result.length){
+      return undefined;
```
fengelniederhammer: Why return undefined in this case? Isn't this rather an error? Is it important to check this here?

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+    timeData.forEach(tt => {
+      const day = tt.day;
+      if (dayMap.has(day)) {
+        const dayGroup = dayMap.get(day);
+        dayMap.set(day, {...dayGroup, ...tt});
+      } else {
+        dayMap.set(day, tt);
+      }
+    });
+    const dayArray = Array.from(dayMap.values()).slice(0, -1); //depending on the day, the latest week just started, so the entropy is calculated as 0 because there are no samples
```
fengelniederhammer: Also according to ChatGPT, this can be simplified to this:
```suggestion
const dayMap = timeData.reduce((map, tt) => {
  const day = tt.day;
  const dayGroup = map[day] || {};
  return {
    ...map,
    [day]: { ...dayGroup, ...tt }
  };
}, {});

const dayArray = Object.values(dayMap).slice(0, -1);
```

Maybe we can also improve the naming of variables here? `tt` should be something understandable, also including `Map` or `Array` usually has no benefit in readability. `Map` can usually be replaced by something like `daysGroupedBy<keys>`. Instead of `Array`, you can simply use the plural (here `days`).

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+    return {dayArray, ticks}
+
+  },[weeklyVariantMutationProportionQuery, selectedGene, variants, deletions]);
+    
+  return data;
+} 
+
+const Plot = ({ plotData, ticks, variants, sequenceType }: PlotProps) => {
+return (
+    <>
```
fengelniederhammer: I think we don't need this fragment, do we?

src/components/NucleotideEntropy.tsx
```diff
@@ -0,0 +1,706 @@
...
+    startPosition: g.startPosition,
+    endPosition: g.endPosition,
+    aaSeq: g.aaSeq,
+    color: chroma.random().darken().hex()
+  };
+}));
+
+
+export const NucleotideEntropy = ({ selector }: Props) => {
+  const [plotType, setPlotType] = useState<string>('time');
```
fengelniederhammer: I think we can improve the typing (and naming) here:
```suggestion
  const [plotType, setPlotType] = useState<'time' | 'position'>('time');
```

src/components/NucleotideEntropy.tsx
```diff
@@ -0,0 +1,706 @@
...
+        plotType={plotType}
+        selectedGenes={selectedGenes}
+        ticks={data.ticks}
+        geneRange={geneRange}
+      />
+    );
+  }
+
+  return (
+    <>
```
fengelniederhammer: This also looks like we don't need that fragment.

src/components/NucleotideEntropy.tsx
```diff
@@ -0,0 +1,706 @@
...
+        value = [{ value: 'All', label: 'All', color: '#353B89' }];
+        break;
+    }
+    setSelectedGenes(value);
+  };
+
+  const controls = (
+    <div className='mb-4'>
+      {/* AA vs. nucs */}
+      <div className='mb-2 flex'>
```
fengelniederhammer: I would guess that we could extract some of those components to reuse them here and in `NucleotideEntropyMultiChart`

src/components/NucleotideEntropy.tsx
```diff
@@ -0,0 +1,706 @@
...
+      : gene.value === 'All'
+      ? posEntropy
+      : posEntropy.filter(g => g.position.includes(gene.value));
+  const sum = filteredPos.map(f => f.entropy).reduce((x, a) => x + a, 0);
+  const count =
+    sequenceType === 'nuc'
+      ? gene.endPosition - gene.startPosition
+      : (gene.value === 'All' ? jsonRefData.genes : jsonRefData.genes.filter(g => g.name.includes(gene.value)))
+          .map(g => g.aaSeq.length)
+          .reduce((x, a) => x + a, 0);
```
fengelniederhammer: This is very hard to read. Could you maybe extract parts of it?

src/components/NucleotideEntropy.tsx
```diff
@@ -0,0 +1,706 @@
...
+              />
+            ))}
+          </LineChart>
+        </ResponsiveContainer>
+      </>
+    );
+  }
+};
+
+const CalculateEntropy = (
```
fengelniederhammer: This should have a name starting with lower case. Only React components should start with a capital letter.

src/components/NucleotideEntropy.tsx
```diff
@@ -0,0 +1,706 @@
...
+        original: pos.original,
+        proportion: remainder,
+      });
+    }
+    pos.proportions = pos.proportions.filter(p => p.proportion > 0);
+  });
+
+
+  //convert proportion to entropy
+  positionGroups.map(p => {
```
fengelniederhammer: Either use `forEach` or return something from the array function in `map`, please.

fengelniederhammer (changes requested):
Please check the code for unnecessary things: There are a couple of fragments, variables and other things that we could get rid of.

Also I would propose to restructure the code:
* create a folder `/src/components/NucleotideEntropy`
* In that folder, create an `index.tsx` that exposes `NucleotideEntropy` and `NucleotideEntropyMultiChart` (either by reexport or by holding the actual implementation). That way, consumers only ever need to know that index.tsx and no other file in the folder.
* Hooks go into separate files in that folder.
* Also extract components into separate files: Either if they are large by themselves, or if they can be reused by multiple other components.
* In general methods should not be longer than 100 lines. Files should also not be much longer than that either.

Also please add unit tests to make sure that your code actually works. In the ideal case, you can split your tests into display and logic tests:
* test the hooks separately. This is where the logic sits. Make sure that the computation is correct. "Given the server returns some specific data, then the hooks computes xyz...".
* test the components separately. Mock the hooks. "Given some specific data to display, then the component shows..."

Also please try to get rid of most of the comments in the code. Comments tend to be outdated and get misplaced, because people forget to update them when changing the code. Some few comments might be justified, but usually one can apply any of:
* The code is already expressive enough and the comment only paraphrases what's written in the code. In that case, simply delete the comment.
* The code explains what's going on in the code. It also paraphrases the code, but the code is more complex, so that it deserves some explanation. In that case, extract the code into a class/method/component (whatever fits) and give it a name such that you don't need the comment, because it's included in the name.

And please use expressive names for all variables. It really helps other people to understand the code a lot faster, if names communicate the purpose of variables.

src/components/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,294 @@
...
+type Variant = {
+  variant: string;
+}
+
+const genes = jsonRefData.genes;
+!genes.map(o => o.name).includes('All') &&
+  genes.push({ name: 'All', startPosition: 0, endPosition: 29903, aaSeq: ''});
+
+function nonNullValues(obj: any) {
+  return Object.entries(obj).filter(([key, value]) => value !== undefined).map(([key, value]) => value).toString();
```
fengelniederhammer: Calling unused variables `_` is a common pattern, if you somehow have to include them, but you don't actually need them:
```suggestion
  return Object.entries(obj).filter(([_, value]) => value !== undefined).map(([_, value]) => value).toString();
```

fengelniederhammer:
You might also want to rebase your branch to the current develop branch. It's quite a few commits behind. :)

src/components/NucleotideEntropy/NucleotideEntropyMultiChart.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const [sequenceType, setSequenceType] = useState<SequenceType>('nuc');
+  const [selectedGene, setSelectedGene] = useState<string>('All');
+  const [includeDeletions, setIncludeDeletions] = useState<boolean>(false);
+
+  const handleDeletions = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setIncludeDeletions(event.target.checked);
+  };
+
+  const exploreUrl = useExploreUrl()!;
+  const variants = (exploreUrl.variants ?? []).map(nonNullValuesToString);
```
fengelniederhammer, line 40: @chaoran-chen Does this particular line make sense? I'm not so sure about it. It seems to work, but it might be quite fragile.

fengelniederhammer (approved):
From my side it looks good now, but it would be good if someone else could double check my refactorings ;)

chaoran-chen (changes requested):
By default, the positions with zero entropy should be included. In fact, I think that we should negate the checkbox to "Exclude positions with zero entropy"

image

Unfortunately, I see that the plot has some performance issues when the positions with zero entropy are included: It takes a long time to load and the zooming is very slow. @fengelniederhammer , do you see a way to improve the performance?

fengelniederhammer:
I'm not sure whether we can do something here. It think it just takes the browser very long to render 30.000 bars. 

The computation is clearly not the problem: When changing the threshold from 0.1 to 0, the data is the same (cached by useMemo). The browser just has to rerender the plot.

Why is it actually useful to see mostly empty bars?

chaoran-chen:
If we exclude the zero entropy bars, the x-axis is not "uniform" anymore. It jumps from 2 to 12, etc.

image

Would it be possible to not actually render the zero entropy positions but just leave the space empty?

chaoran-chen (approved):
As further performance improvements seem difficult at the moment, I'll merge it.

Thank you very much, @philipschulz and @fengelniederhammer for the great work!

PR #789, opened 2023-04-26
Tooltips of wastewater plots are not completely visible

### LINKED ISSUE #788: Tooltips of wastewater plots are not completely visible
On https://cov-spectrum.org/stories/wastewater-in-switzerland, the lower part of the tooltips is not visible.

image

JonasKellerer:
Resolves #788

This is a first draft of the fix. Currently the position of the tooltip is raised by 80. This works on my screen, and also with smaller window sizes. It works also on the mobile emulator, but there is not the problem, that all is shown, but that too much is shown. An increase of the zindex did not fix the issue. There might be more to it, but I could not find it, why this does not work.

If we get more variants, I think we can go the route to reduce the spacing between the lines, or to migrate to a two column view.

Sadly, I could not write a tests, since I was unable to get the tooltip to show. I could not find an element, to hover over, which worked.

So, feedback welcome :)

### DIFF
--- a/src/components/PackedGrid/PackedGrid.tsx
+++ b/src/components/PackedGrid/PackedGrid.tsx
@@ -72,7 +72,7 @@ export const PackedGrid = ({ children, maxColumns }: Props) => {
               key={child.key === null ? `child-${cell.index}` : `child-around-${child.key}`}
               node={portalNodes[cell.index]}
             >
-              <div style={{ width: cell.width, overflow: 'hidden' }}>{child.props.children}</div>
+              <div style={{ width: cell.width }}>{child.props.children}</div>
             </InPortal>
           );
         })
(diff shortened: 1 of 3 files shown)


chaoran-chen (approved):
Thanks, @JonasKellerer! It also works well for me! We are likely going to redo the wastewater plots and pages at some point this year but for the moment, that should suffice.

JonasKellerer:
With the new feedback in the issue #788 I found out, that the tooltip is croped by the element `PackedGrid` through the style option `overflow: 'hidden'`. If I set it to `visible`, the tooltip is shown correctly. I am currently investigating if we could losen this restriction on the grid.

fengelniederhammer (approved):

PR #794, opened 2023-05-09
updated variants colors

mcarrara-bioinfo:
After including XBB.1.5, XBB.1.16 and XBB.1.9 as well as excluding the general XBB*, I updated the colors for the curves according to the colors suggested by covariants


chaoran-chen (approved):
thanks for the update!

PR #795, opened 2023-05-15
Curves visualization improvements

### LINKED ISSUE #791: Curves visualization improvements
As part of NEXUS I am supporting the cEVO team in the publication of the [Swiss wastewater curves](https://cov-spectrum.org/stories/wastewater-in-switzerland) on the website. We received pointers from the BAG on user experience regarding the plots of the curves:
- Time frame: Each sampling location currently shows only the time frame where data is available, i.e. the origin of the x-axis is the first available sampling point for the specific location and the axis shows up to the latest available sampling point for the location. To improve user experience when comparing sampling locations, the x-axis should refer to the full time frame available throughout all locations, i.e. the x-axis origin should be the earliest sampling date available and the axis should show up to the latest available sampling point between all locations. Below is an example of the expected result for two sampling locations:
image
- Ticks: Currently the X-axis has a total of 3 ticks representing `<day>.<month>`. Due to the large time frame available in some sampling locations it was suggested to increase clarity by increasing the total amount of ticks shown and by including the year as part of the tick label.

JonasKellerer:
Resolves: #791

This is a first draft and work in progress (tests and refactoring necessary). Do we already have the wanted functionality @chaoran-chen?
What should be displayed, when no data is available?

This is how the selector could look like:
image

This is how partially missing data could look like:
image

This is how no data, or only one data point could look like:
image


chaoran-chen:
Thanks, @JonasKellerer, it seems great to me! (Just showing "no data" sounds like a good solution to me.)

@mcarrara-bioinfo, could you please take a look (just click on "Visit Preview" above) and let us know whether it is as you expected?

mcarrara-bioinfo:
> @mcarrara-bioinfo, could you please take a look (just click on "Visit Preview" above) and let us know whether it is as you expected?

Thank you very much for the quick reaction to the issue. Everything looks great!
Would adding more ticks make the x-axis too busy in your opinion? It's one of the pointers we receive, but I do realise that the detailed pop-up at mouse-over helps with pinpointing dates of interest (and a feature to select the desired time frame as I see in the first picture) are already a solution.
I completely agree with the proposal on how to show no data available and single time points.

JonasKellerer:
I think that adding more ticks makes the x-axis too busy. Especially if we consider, that the website will be displayed on various window sizes. So I would suggest, leaving it like this and if the user feedback persists we have to think about it.

src/helpers/format-data.ts
```diff
@@ -0,0 +1,11 @@
+export function formatCiPercent(ci: [number, number]): string {
+  return `[${formatProportion(ci[0])} - ${formatPercent(ci[1])}]`;
+}
+
+export function formatPercent(value: number, digitsAfterDot?: number): string {
+  return formatProportion(value, digitsAfterDot) + '%';
+}
+
+export function formatProportion(value: number, digitsAfterDot?: number): string {
+  return (Number(value) * 100).toFixed(digitsAfterDot ?? 2);
```
fengelniederhammer:
```suggestion
export function formatProportion(value: number, digitsAfterDot: number = 2): string {
  return (Number(value) * 100).toFixed(digitsAfterDot);
```

src/models/wasteWater/WasteWaterLocationTimeChartTooltip.tsx
```diff
@@ -0,0 +1,50 @@
+import { TooltipProps } from 'recharts';
+import { NameType, ValueType } from 'recharts/types/component/DefaultTooltipContent';
+import { formatDate } from './WasteWaterTimeChart';
+import React from 'react';
+import { formatCiPercent, formatPercent } from '../../helpers/format-data';
+import { deEscapeValueName } from './RechartsKeyConversion';
+
+export const WasteWaterTooltip = ({ active, payload }: TooltipProps<ValueType, NameType>) => {
+  if (active && payload && payload.length > 0) {
```
fengelniederhammer: Nitpicking: Usually it's easier to read with the shorter `return` first:
```
if (...) {
  return null;
}
```

src/models/wasteWater/__tests__/WasteWaterLocationTimeChart.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    const dateRange = {
+      dateFrom: globalDateCache.getDay('2020-01-01'),
+      dateTo: globalDateCache.getDay('2020-01-02'),
+    };
+
+    const variants = getWasteWaterLocationTimeChartProps(datesOfData, variantNames);
+
+    render(
+      <div style={{ width: 500, height: 500 }}>
```
fengelniederhammer: Why do we need that `div` in this test, but not in the other test?

src/models/wasteWater/story/WasteWaterSamplingSitesHooks.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export function useWasteWaterData(): WasteWaterDataWithLocation[] | undefined {
+  const [wasteWaterData, setWasteWaterData] = useState<WasteWaterDataset | undefined>(undefined);
+  useEffect(() => {
+    getData({
+      country: 'Switzerland',
+    }).then(dataset => dataset && setWasteWaterData(dataset));
+  }, []);
+
+  const locationData = useMemo(() => {
```
fengelniederhammer: IntelliJ suggests to inline this:
```suggestion
  return useMemo(() => {
```

fengelniederhammer (approved):
Looks good to me

src/models/wasteWater/__tests__/WasteWaterLocationTimeChart.test.tsx
JonasKellerer: Actually we dont. It was a leftover before the ResponsiveContainer was mocked.

PR #820, opened 2023-06-22
added xbb.2.3 to color palette

mcarrara-bioinfo:
BAG raised XBB.2.3 as a new variant of interest.
Included the variant to the color palette


chaoran-chen (approved):
Thanks!

PR #843, opened 2023-08-19
fix(#842): pass LapisDataVersion as prop into footer

### LINKED ISSUE #842: BUG: Dataversion is often $NOW rather than datetime
I noticed that I often see the current time rather than the dataversion:
image

Looking at the source code (but not setting up cov-spectrum itself) I think I've figured out the root cause: you don't make sure to fetch the current datetime before setting it.

This is the react code that displays things in the footer:
https://github.com/GenSpectrum/cov-spectrum-website/blob/6412c722bafb1d3136f22545671706ae5417c40c/src/App.tsx#L253

`getCurrentLapisDataVersionDate()` notably _doesn't_ fetch Lapis' datetime, it just checks what the current datetime is: 
https://github.com/GenSpectrum/cov-spectrum-website/blob/6412c722bafb1d3136f22545671706ae5417c40c/src/data/api-lapis.ts#L66-L68

Unfortunately, rather than throwing an error, `dayjs(undefined).locale("en").calendar()` returns the current datetime making this bug hard to spot.

I don't know enough React to fix the bug - one could potentially just throw an `fetchLapisDataVersionDate()` into getcurrentLapisDataVersionDate(), maybe like so:

```ts
function getCurrentLapisDataVersionDate() {
  if (currentLapisDataVersion === undefined) {
    fetchLapisDataVersionDate();
  }
  return currentLapisDataVersion !== undefined
    ? dayjs.unix(currentLapisDataVersion).toDate()
    : undefined;
}
```

I have no clue why this bug shows only _sometimes_. Maybe because of stochasticity in page load behaviour and async stuff.

Here's a repro script:

```js
const dayjs = require("dayjs");
//import dayjs from 'dayjs' // ES 2015

var calendar = require("dayjs/plugin/calendar");
dayjs.extend(calendar);

let currentLapisDataVersion = undefined;
const HOST = "https://lapis.cov-spectrum.org/open/v1"; // Replace with the appropriate HOST
const ACCESS_KEY = ""; // Replace with the appropriate ACCESS_KEY or remove if not needed

const get = async (endpoint, signal, omitDataVersion = false) => {
  let url = HOST + endpoint;
  if (currentLapisDataVersion !== undefined && !omitDataVersion) {
    url += "&dataVersion=" + currentLapisDataVersion;
  }
  const res = await fetch(url, {
    method: "GET",
    signal,
  });
  if (res.status === 410) {
    window.location.reload();
  }
  return res;
};

async function fetchLapisDataVersionDate(signal) {
  let url = "/sample/info";
  if (ACCESS_KEY) {
    url += "?accessKey=" + ACCESS_KEY;
  }
  const res = await get(url, signal);
  if (!res.ok) {
    throw new Error("Error fetching info");
  }
  const info = await res.json();
  currentLapisDataVersion = info.dataVersion;
}

function getCurrentLapisDataVersionDate() {
  if (currentLapisDataVersion === undefined) {
    fetchLapisDataVersionDate();
  }
  return currentLapisDataVersion !== undefined
    ? dayjs.unix(currentLapisDataVersion).toDate()
    : undefined;
}

// Example usage (make sure you adjust constants like HOST and ACCESS_KEY appropriately):
// Print the current data version date
console.log(dayjs(getCurrentLapisDataVersionDate()).locale("en").calendar());

// Fetch the current data version date
fetchLapisDataVersionDate().then(() => {
  // Print the current data version date
  console.log(dayjs(getCurrentLapisDataVersionDate()).locale("en").calendar());
});
```

This shows the following for me:
```bash
❯ node bug.js
Today at 1:42 AM
07/23/2023
```

corneliusroemer:
resolves #842

Make sure that the lapisDataVersion is passed as a prop, rather than relying on a global variable

Mirror fetchNextcladeDatasetInfo pattern for implementation

Get rid of strange code remnant baseLocation.ts that seemd to have nothing to do with the baseLocation

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,8 +12,7 @@ import { WasteWaterLocationPage } from './models/wasteWater/story/WasteWaterLoca
 import StoriesOverview from './stories/StoriesOverview';
 import StoryRouter from './stories/StoryRouter';
 import { useExploreUrl } from './helpers/explore-url';
-import dayjs from 'dayjs';
-import { fetchNextcladeDatasetInfo, getCurrentLapisDataVersionDate } from './data/api-lapis';
+import { fetchNextcladeDatasetInfo, fetchLapisDataVersion } from './data/api-lapis';
 import { sequenceDataSource } from './helpers/sequence-data-source';
 import { ExternalLink } from './components/ExternalLink';
 import styled from 'styled-components';
@@ -59,6 +58,7 @@ export const App = () => {
   const isSmallScreen = width !== undefined && width < 768;
 
   const { data: nextcladeDatasetInfo } = useQuery(() => fetchNextcladeDatasetInfo(), []);
+  const { data: lapisDataVersion } = useQuery(() => fetchLapisDataVersion(), []);
 
   const isChatPage = useLocation().pathname === '/chat';
   const showFooter = !hideHeaderAndFooter && !isChatPage;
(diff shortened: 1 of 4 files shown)


JonasKellerer (approved):
Nice fix. Thank you.

chaoran-chen:
Thanks, @corneliusroemer!

PR #848, opened 2023-08-28
BA.2.86 colour in constants.ts

DrYak:
- No official colour in covariants.org
- BA.2.86 deconvolution is active for Wastewater (though curves close to 0 for now).


chaoran-chen (approved):
Thanks!

PR #879, opened 2023-11-07
Update constants.ts to follow covariants.org definitions

mcarrara-bioinfo:
Removing improvised colors and including the most recent covariants.org definitions


chaoran-chen (approved):
Thanks!

Issue #735, opened 2023-02-21
Console shows warnings, when clicking on covSpectrum logo


JonasKellerer:
When on the [explore page](http://localhost:3000/explore/Germany/AllSamples/Past6M/variants?nextcladePangoLineage=ba.2.75*&) and then clicking on the covSpectrum logo in the top left, the console throws many warnings. It seems not to influence the behavior of the page. But one could have a look at it at some point.

Selecting countries in the header does not throw these warnings.

![image](https://user-images.githubusercontent.com/122305307/220321597-30c0bf80-79dd-49ed-b36b-ef2b1cc16cc7.png)

PR #883, opened 2023-11-18
feat: improve Header links

wszczawinski:
- improve Logo link path,
- use Link component instead of href in Header (Link prevents full page refresh)
- logoClick

#735


fengelniederhammer (approved):
Thanks for the contribution, looks good to me :)

PR #901, opened 2024-01-11
Use locale for time conversion for wastewater, not iso time string

### LINKED ISSUE #896: Wastewater curves do not display all available dates in selected plots
I have just updated the curves for the "stories" page about the wastewater in Switzerland.

The database reports that for Chur and Lausanne there are data point up to and including December 03, while for Altenrhein up to and including December 04.
However, with mouse-over at the end of the curves, we can only see data points up to December 02 for Chur and Lausanne, and up to December 03 for Altenrhein.

JonasKellerer:
Resolves #896 

Now the time string is no longer converted from the ISO time, but the locale, where the user looks at the data. This is now the same way as all the others plots on cov-spectrum.

### DIFF
--- a/src/models/wasteWater/types.ts
+++ b/src/models/wasteWater/types.ts
@@ -7,8 +7,6 @@ export const WasteWaterSelectorSchema = zod.object({
   location: zod.string().optional(),
 });
 
-export type WasteWaterSelector = zod.infer<typeof WasteWaterSelectorSchema>;
-
 export type WasteWaterRequest = {
   country: string;
 };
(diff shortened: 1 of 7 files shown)


src/models/wasteWater/types.ts
```diff
@@ -7,8 +7,6 @@ export const WasteWaterSelectorSchema = zod.object({
   location: zod.string().optional(),
 });
 
-export type WasteWaterSelector = zod.infer<typeof WasteWaterSelectorSchema>;
```
JonasKellerer, line 10: I removed these types, since they are not used.

chaoran-chen (approved):
Great, thank you!

PR #937, opened 2024-02-28
Do not merge

JonasKellerer:
This should only provide a vercel deployment for load testing.


JonasKellerer:
We addressed the load testing in https://github.com/GenSpectrum/LAPIS-SILO/pull/301 . We no longer need this PR

PR #947, opened 2024-03-15
fix: reload when SILO is in maintenance mode

### LINKED ISSUE #943: EDIT: Resolved as soon as I submitted this -- All queries fail
EDIT: This was a temporary issue and is working for me again. Feel free to close if this doesn't seem important. 

I believe this started for me sometime between March 7 16:45 EST and 17:15 EST. 

All queries, having tried on both a PC and phone, result in the loading icon running forever. The UI stays in tact and is interactive though. 
image

image
Failed to load resource: the server responded with a status of 503 ()
instrument.ts:124 Error: Error fetching info
    at api-lapis.ts:83:11
    at p (regeneratorRuntime.js:45:16)
    at Generator.<anonymous> (regeneratorRuntime.js:133:17)
    at Generator.next (regeneratorRuntime.js:74:21)
    at o (asyncToGenerator.js:3:20)
    at l (asyncToGenerator.js:22:9)
(anonymous) @ instrument.ts:124
Show 1 more frame
Show less
2instrument.ts:124 DOMException: The user aborted a request.

fengelniederhammer:
resolves  #943

This tries to fix the following scenario:
Given you're on the website
Given SILO now goes to maintenance mode (returning 503) When I navigate on the website
Then I should see the maintenance page

Previously, we only checked the maintenance mode when initially loading the page.

I reproduced this locally, there the fix works.

### DIFF
--- a/src/data/api-lapis.ts
+++ b/src/data/api-lapis.ts
@@ -36,7 +36,11 @@ const ACCESS_KEY = process.env.REACT_APP_LAPIS_ACCESS_KEY;
 
 let currentLapisDataVersion: number | undefined = undefined;
 
-export const get = async (endpoint: string, signal?: AbortSignal) => {
+export const get = async (
+  endpoint: string,
+  signal?: AbortSignal,
+  options: { skipMaintenanceCheck?: boolean } = {}
+) => {
   let url = `${HOST}/sample${endpoint}`;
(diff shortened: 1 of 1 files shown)


JonasKellerer (approved):
LGTM

PR #985, opened 2024-05-21
feat: disable the chat

chaoran-chen:
This disables the chat. It removes links to the `/chat` page. The page itself still exists (to not break existing links and bookmarks) and informs the visitor that the chat is not available.

image


fengelniederhammer (approved):
LGTM :+1:

PR #986, opened 2024-05-21
feat: disable estimated cases and reproduction number plots

chaoran-chen:
I disabled the estimated cases and reproduction number plots because we don't have the case data at the moment.


src/pages/FocusSinglePage.tsx
```diff
@@ -56,6 +56,7 @@ import { AnalysisMode } from '../data/AnalysisMode';
 import { NucleotideEntropy } from '../components/NucleotideEntropy/NucleotideEntropy';
 // Due to missing additional data, we are currently not able to maintain some of our Swiss specialties.
 const SWISS_SPECIALTIES_ACTIVATED = false;
+const CASES_DISABLED = true;
```
fengelniederhammer:
```suggestion
const CASES_ENABLED = false;
```

To prevent the double negation `!disabled`

fengelniederhammer (approved):
LGTM

PR #1003, opened 2024-06-27
feat: remove deep hospitalization-death page

### LINKED ISSUE #987: Improve load performance by removing whatever requires the single huge 11MB request
On an 15MB network connection, the page load (world) is extremely slow.

Looking at network tab, there's an obvious culprit, a single 11MB request:
https://lapis.cov-spectrum.org/gisaid/v2/sample/aggregated?host=Human&accessKey=9Cb3CqmrFnVjO3XCxQLO6gUnKPd&fields=division%2Cage%2Csex%2Chospitalized%2Cdied

This looks like to me it's only used by the age/sex/hospitalized/died panels on covSpectrum which I've never used, and I think most people don't use it so the easiest fix would be to remove the panel that requires that data.

Also, the request seems to not be cached at all, so it slows _every_ page load

Brave Browser 2024-05-23 14 38 03

chaoran-chen:
Resolves #987

I haven't removed the plot components entirely (i.e., they are currently dead code) to make it a bit easier in the (unlikely) case that we want to use them again. As this code base is not actively developed at the moment (we will hopefully replace it with the new dashboards at some point), it shouldn't be a problem to have the code around for now.


JonasKellerer (approved):
LGTM.

Do we want to add a comment on the dead code with a ticket? So we know either to leave it there, or to remove it later on?

PR #1088, opened 2025-06-18
Update variant colors in constants.ts

kirschen-k:
Added 2 variants XFG and NB.1.8.1 to the list with the corresponding colors from covariants.org. Corrected the colors for JN.1 and XEC to be following the ones from covariants.org.


chaoran-chen (approved):
Thanks!

PR #1110, opened 2025-07-24
chore(deps): update various dependencies

chaoran-chen:
This is updating a first set of dependencies that should be rather unproblematic. I'll submit PRs for more complicated updates (MUI, Recharts and React) separately.


anna-parker (approved):
approving as all the code changes affect tests (@chaoran-chen is there a standard way to test these updates for cov-spectrum?)

chaoran-chen:
@anna-parker, we unfortunately don't have a very high test coverage in cov-spectrum, so it relies a lot on clicking/typing through the preview and look for breaks. This PR's updates are relatively harmless, so I'm not very worried but with the upcoming ones, we need to check very carefully.

PR #1114, opened 2025-07-26
chore(deps): bump @mui/material to 6 and @mui/x-data to 7

chaoran-chen:


src/components/chat/__tests__/IncomingResponseMessage.test.tsx
```diff
@@ -54,25 +54,28 @@ describe('IncomingResponseMessage', function () {
...
-    };
-
-    renderIncomingResponseMessage(message);
-
-    expect(screen.getByText('textBeforeData')).toBeInTheDocument();
-    expect(screen.getByText('textAfterData')).toBeInTheDocument();
-    expect(screen.getByText('Proportion (%)')).toBeInTheDocument();
-    expect(screen.getByText('12.35')).toBeInTheDocument();
-  });
+  // The IncomeResponseMessage component – along with the entire ChatPage – is currently not under use, so
```
anna-parker, line 57: so does this mean that the ChatPage will stop working with this update or that the tests are broken? Does the ChatPage still exist on cov-spectrum? Maybe we want to delete it first to not have a broken page

src/widgets/__tests__/__snapshots__/EstimatedCasesChart.test.tsx.snap
```diff
@@ -17,13 +17,13 @@ exports[`<EstimatedCasesChart> dataset0 renders correctly 1`] = `
         2021-12-29
       </b>
       <div
-        className="MuiFormGroup-root css-dmmspl-MuiFormGroup-root"
+        className="MuiFormGroup-root css-1mac7pt-MuiFormGroup-root"
```
anna-parker, line 20: I see that all these IDs have changed but I don't see what they are regarding... what is `css-1mac7pt-MuiFormGroup-root` labeling?

src/components/chat/__tests__/IncomingResponseMessage.test.tsx
chaoran-chen: The code here is unused as there is no actual chat feature at the moment but instead we show at https://cov-spectrum.org/chat this:

image

I created an issue to remove the code (#1128).

src/widgets/__tests__/__snapshots__/EstimatedCasesChart.test.tsx.snap
chaoran-chen: In this case it is labeling the the checkbox for "Log scale".

chaoran-chen:
I've used this now for a while as part of #1119. I'm mostly confident that this is alright and will therefore go ahead and merge it now.

PR #1143, opened 2025-09-18
fix: replace react-boostrap collapse with tailwind collapse

### LINKED ISSUE #1142: Wastewater Discontinued Sites Plots Not Displaying in Collapsible Sections
On the Wastewater Story page, the discontinued sites are shown in collapsible sections. When expanding a section, the plot(s) for the discontinued site(s) sometimes briefly appear, but then disappear, leaving the expander open but blank (white). The expander always opens to the correct size, as if the plots are rendered, but the content is not visible.

I've tried to resolve it myself, but this appears to be a more intricate React UX issue.

**What I’ve Tried** (consider with suspicion)
- Checked data and date logic: The data and date ranges are correct; this is not a data issue.
- Memoized date range selector: Used [useMemo](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to avoid recreating the date range selector object on every render.
- Conditional rendering: Only renders the plot component when the expander is open.

**Observations**
- The expander opens to the correct height for the plot, but the plot is not visible.
- Sometimes the plot flashes briefly before disappearing.
- The issue is not present for non-discontinued sites (plots render fine outside the collapsible).
- The problem started recently, but no obvious code change seems to have caused it.

Reported by @chaoran-chen

fhennig:
resolves #1142 

I'm not entirely sure what the issue was before with the `react-boostrap` component. I actually suspect that maybe its CSS classes were interfering with the tailwind classes?

Tailwind can also do collapsing, so I got rid of the `Collapse` component and used simple CSS classes instead.

https://github.com/user-attachments/assets/f5cf262a-8744-4cfd-ac46-461add7225c9

### DIFF
--- a/src/models/wasteWater/story/WasteWaterStoryPage.tsx
+++ b/src/models/wasteWater/story/WasteWaterStoryPage.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useState } from 'react';
-import { Collapse } from 'react-bootstrap';
 import { ExternalLink } from '../../../components/ExternalLink';
 import { isDiscontinuedSite, WasteWaterSamplingSites } from './WasteWaterSamplingSites';
 import { discontinuedSites } from '../constants';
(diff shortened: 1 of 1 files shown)


chaoran-chen (approved):

gordonkoehn:
Looks great me!

PR #1159, opened 2025-11-12
feat: mutations over time: use new endpoint

fhennig:
WIP

### DIFF
--- a/src/data/api-lapis.ts
+++ b/src/data/api-lapis.ts
@@ -356,7 +356,7 @@ export async function fetchMutationsOverTime(
   // Add accessKey if available
   let endpointWithParams = `/${endpoint}`;
   if (ACCESS_KEY) {
-    endpointWithParams += '&accessKey=' + (await _getCurrentAccessKey());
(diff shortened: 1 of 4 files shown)


src/data/api-lapis.ts
```diff
@@ -276,6 +332,91 @@ function getInsertionEndpoint(sequenceType: SequenceType): string {
...
+    // Try to parse as number if it looks like a number
+    const numValue = Number(value);
+    if (!isNaN(numValue) && value !== '') {
+      filters[key] = numValue;
+    } else {
+      filters[key] = value;
+    }
+  });
+  return filters;
+}
```
fhennig: This function makes it really easy for us to reuse all the code related to the URL params, but also maybe the number conversion could be problematic. I think we should be fine, but wanted to point it out anyways.

fhennig: Ah damn, I think maybe this is actually an issue for lists etc.

chaoran-chen (approved):

PR #1177, opened 2026-04-02
chore: use Node 24 to build the app

### LINKED ISSUE #1130: Update Node.js
We're still using Node 18 which is end-of-life since May this year. Let's update to Node 22 or 24.

The Dockerfiles and GitHub actions need to be updated.

fengelniederhammer:
resolves #1130


fengelniederhammer:
I built the Docker image locally and it was working.

chaoran-chen (approved):

